PR #63, opened 2021-05-28
fix: 60 issue - show timeline icon when sidepanel is open

adriana-carmo:
Fix: 1) timeline icon should not show when side panel is open
timeline icon 

2) should never show when viewing a single tree link (like ?treeid=12312 or ?tree_name=test-name)


dadiorchen (commented):
@adriana-carmo thank you!

PR #91, opened 2021-08-08
feat: greenstand logo is a link

mmayeda:


dadiorchen:
@mmayeda thanks for your first PR!
There is a tiny problem on this PR, that is: 
The rule is: the app would try to load the map with different logo:
https://github.com/Greenstand/treetracker-web-map-client/blob/d5f35916ce18478610e6a2be5988fefbf878de97/src/App.js#L503-L508
You know there are a lot of different organization on Greenstand's map, so the logo is not fixed, so I guess the rule is: if the logo is greenstand, we need to add the link, if not, just leave it empty, in the future, we will make this constumizable, so our other organizations could be able to set their own link here.

dadiorchen:
Thanks for you work!

mmayeda:
Ah I should have noticed that, thanks for taking a look!  Should only link now when the logo source references the greenstand one in the project

src/components/Logo.cy.js
```diff
@@ -0,0 +1,24 @@
...
+    const testSrc = require("../images/logo_floating_map.svg");
+    mount(<Logo logoSrc={testSrc} logoLoaded={true}/>)
+    cy.get("a > img")
+      .should("be.visible")
+      .should("have.attr", "src", testSrc);
+    cy.get("a")
+      .should("exist")
+      .should("have.attr", "href", "https://greenstand.org/")
+  })
+});
```
dadiorchen, line 24: I like your trying on this unit test by Cypress :D

src/components/Logo.js
```diff
@@ -0,0 +1,50 @@
...
+//    [theme.breakpoints.down("sm")]: {
+//      transform: "translate(0, 0)",
+//      opacity: 1,
+//    },
+  },
+}));
+
+function Logo(props){
+  const classes = useStyles();
+  const isGreenstandLogo = props.logoSrc === require("../images/logo_floating_map.svg")
```
dadiorchen, line 37: Yes, this is maybe the best way to check it.

src/App.js
```diff
@@ -573,11 +574,10 @@ function App() {
...
       </Fade>
-      <div className={`${classes.logo} ${logoLoaded?classes.logoLoaded:""}`}>
-        <a href="https://greenstand.org/" target="_blank" rel="noopener noreferrer">
-          <img alt="logo" src={logoSrc} />
-        </a>
-      </div>
+      <Logo
+        logoLoaded={logoLoaded}
+        logoSrc={logoSrc}
+      />
```
dadiorchen, line 580: Just for discussion, how about putting the load process into the Logo, I don't if this idea has any problem, just thinking, if you look at the logo as a component/class, then the way to load the logo is might be an appropriate thing this class should KNOW.

dadiorchen (approved):

src/App.js
mmayeda: I think it makes a lot of sense for the logo component to own that logic!  I'll adjust the PR

mmayeda: After looking at this more, I think keeping the logo loading logic outside of the component leaves the component with a single responsibility and superior testability.  Let me know if you agree and I can merge

PR #97, opened 2021-08-22
test: fix map test

mmayeda:


dadiorchen:
@mmayeda thanks for your great contribution for this org

PR #115, opened 2021-09-15
fix: zoom on mobile issue

### LINKED ISSUE #98: Can not pan the map in mobile device.
https://www.loom.com/share/1ae6fab6f524426d895dd15d6c8215fe

cpaule1811:
Resolves #98 

zoom-issue

Hi, 

The issue was due to the logo div's height not being static so the click event on zoom couldn't propagate because the div was blocking it.  

By setting the height of the logo div to 50px, it no longer prevents zoom click event from happening

https://user-images.githubusercontent.com/80873472/133403625-b7da83a7-b419-42ae-8149-0fcf741fbd6f.mp4

### DIFF
--- a/src/components/Logo.cy.js
+++ b/src/components/Logo.cy.js
@@ -20,6 +20,5 @@ describe("Logo", () => {
     cy.get("a")
       .should("exist")
       .should("have.attr", "href", "https://greenstand.org/")
-      cy.viewport(800, 500)
   })
 });
--- a/src/components/Logo.cy.js
+++ b/src/components/Logo.cy.js
@@ -20,5 +20,6 @@ describe("Logo", () => {
     cy.get("a")
       .should("exist")
       .should("have.attr", "href", "https://greenstand.org/")
+      cy.viewport(800, 500)
   })
 });
--- a/src/components/Logo.js
+++ b/src/components/Logo.js
@@ -5,7 +5,6 @@ const useStyles = makeStyles(theme => ({
   logo: {
     userSelect: "none",
     position: "absolute",
-    height: 50,
     right: 60,
     bottom: 20,
     opacity: 0,
@@ -15,11 +14,13 @@ const useStyles = makeStyles(theme => ({
       maxWidth: 250,
     },
     [theme.breakpoints.down("sm")]: {
+      pointerEvents: "none",
       transform: "translate(0, -20px)",
       right: 10,
       top: 10,
       "& img": {
         width: "45vw",
+        pointerEvents: "auto",
       },
     },
   },
--- a/src/components/Logo.js
+++ b/src/components/Logo.js
@@ -5,6 +5,7 @@ const useStyles = makeStyles(theme => ({
   logo: {
     userSelect: "none",
     position: "absolute",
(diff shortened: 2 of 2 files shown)


dadiorchen:
@cpaule1811 awesome! this is a big problem for us, I didn't realize it is just because of this funny reason :D

src/components/Logo.js
```diff
@@ -5,6 +5,7 @@ const useStyles = makeStyles(theme => ({
   logo: {
     userSelect: "none",
     position: "absolute",
+    height: 50,
```
dadiorchen, line 8: Yes, set the height is a good way to solve the problem, but I am thinking mayge the css property: user-event (I forgot the name) would be another approach, basically, with this property, the whole layer/div become un-interactable for the user.   (pointer-events ?)

dadiorchen: @mmayeda also I think if we remove the user interaction on the logo (for the mobile device), than we need to re-think the logo link, right?

cpaule1811: Yeah, so we can set pointerEvent: "none", however as you mentioned this would take away user interaction from the logo but it would fix the zoom controls issue if the logo interaction isn't important.

On a side note, setting height to 50 messes with the positioning on a large screen device, however if the height is closer to the actual logo container's height that won't be an issue. So 30 is probably a more appropriate height if we do choose to go down that route for this fix.

dadiorchen: I think for mobile device, we can remove the logo action? so using the material-ui breaking-point CSS property, we could set the pointerEvent on the mobile device?

dadiorchen: @mmayeda what do you think, I thought on the mobile device, it's not so necessary to click the logo, considering the screen is small.

mmayeda: I think we can set the pointer events to "none" on the parent div and add pointer events "auto" to the link element to keep the link functionality.  But I'm totally fine with removing the link functionality on the smaller screen too if it becomes a source of frustration by the way of users accidentally clicking it while trying to pan around on the smaller screen

dadiorchen: @mmayeda good idea, also, I agree, let's disable the logo action on the mobile device, @cpaule1811 can you make this change?

cpaule1811: No problem.

src/components/Logo.cy.js
```diff
@@ -20,5 +20,6 @@ describe("Logo", () => {
     cy.get("a")
       .should("exist")
       .should("have.attr", "href", "https://greenstand.org/")
+      cy.viewport(800, 500)
```
mmayeda, line 23: I don't think adjusting the viewport after making assertions buys us anything here.  Can we remove this line?

mmayeda (approved):
Thanks LGTM!

dadiorchen:
Thank you @cpaule1811 for resolve this serious bug, a good improvement for us, I didn't expect it to be because of this reason, I thought it because some tricky bug with leaflet!

dadiorchen:
I will release this version ASAP

PR #129, opened 2021-09-18
Issue100/new logo

mmayeda:
From the design mocks it looks like we are giving up the old logo on the map for the static greenstand logo in the upper left, so I repurposed the existing Logo component instead of creating a new one.  Let me know if that assumption is incorrect
logo_component


dadiorchen:
Perfect!

PR #140, opened 2021-09-23
docs: correct readme to run next

dadiorchen:
Maybe `dev` is better than `next start`? 
When I run next start, the terminal throws error complaining no .next found, so I run `npm run dev`


dadiorchen:
@cpaule1811

cpaule1811:
Okay, I will update the read me. 

Would you like me to update the cypress part of the docs while i'm at it because that will work a little differently from now as well?

cpaule1811:
Sorry just looked at your commit. Yeah, change it to npm run dev. npm start will only work if you npm run build first.

cpaule1811 (approved):

PR #141, opened 2021-09-23
Issue102

dadiorchen:
@cpaule1811 I made the home page following the design, it goes well with the next, but I guess we are still rendering on client-side, right? see my comment below.
image


src/pages/index.js
```diff
@@ -1,6 +1,6 @@
 import dynamic from 'next/dynamic';
 
-const DynamicApp = dynamic(() => import('../App'), { ssr: false });
+const DynamicApp = dynamic(() => import('../Home'), { ssr: true });
```
dadiorchen, line 3: When I change ssr to true, I got the window object not found error, just like what we have discussed, right? 
 For the Home page, we want to render the content on the left on server-side, and render the map on the right on client-side, so we need to resolve the window-importing problem?
Any idea on this?

```
A codemod is available to fix the most common cases: https://nextjs.link/codemod-ndc
info  - ready on http://localhost:3000
ReferenceError: window is not defined
    at /Users/deanchen/work/greenstand/treetracker-web-map-client/node_modules/leaflet/dist/leaflet-src.js:227:1
9
    at /Users/deanchen/work/greenstand/treetracker-web-map-client/node_modules/leaflet/dist/leaflet-src.js:7:66
    at Object.<anonymous> (/Users/deanchen/work/greenstand/treetracker-web-map-client/node_modules/leaflet/dist/
leaflet-src.js:10:2)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.leaflet (/Users/deanchen/work/greenstand/treetracker-web-map-client/.next/server/pages/index.js:40
7:18)
    at __webpack_require__ (/Users/deanchen/work/greenstand/treetracker-web-map-client/.next/server/webpack-runt
ime.js:33:42)
    at eval (webpack-internal:///./src/App.js:2:65)
    at Object../src/App.js (/Users/deanchen/work/greenstand/treetracker-web-map-client/.next/server/src_Home_js.
js:40:1)
    at __webpack_require__ (/Users/deanchen/work/greenstand/treetracker-web-map-client/.next/server/webpack-runt
ime.js:33:42)
    at eval (webpack-internal:///./src/Home.js:5:62)
    at Object../src/Home.js (/Users/deanchen/work/greenstand/treetracker-web-map-client/.next/server/src_Home_js
.js:51:1)
wait  - compiling...
warn  - ./src/components/Loader.js
Anonymous function declarations cause Fast Refresh to not preserve local component state.
Please add a name to your function, for example:
```

dadiorchen:
By the way, amazing experience coding with next!

cpaule1811:
Yes, everything we want server side rendered will have to stay out of App.js. So we might even want to consider taking the SidePanel component and refactoring so it works out side of the map. It may also be worth changing the name of App.js to MapComponent.js or something so it does not get confused with _app.js.

Cypress component testing should work with npx cypress open-ct
and integration tests for pages can be placed in cypress/integration folder and ran with npx cypress open. 
Unless you have found a better way? 

It would be a good idea to put <Navbar/> and anything else to do with page layout in layout.js. That way we can avoid repeating of code and unnecessary renders. 

Yeah Next is a pleasure to work with and quite intuitive for the most part.

src/Home.js
```diff
@@ -0,0 +1,115 @@
...
+import Grid from "@material-ui/core/Grid";
+import Typography from "@material-ui/core/Typography";
+import Button from "@material-ui/core/Button";
+import Box from "@material-ui/core/Box";
+import Paper from "@material-ui/core/Paper";
+import Navbar from "./components/Navbar";
+import log from "loglevel";
+import dynamic from "next/dynamic";
+
+const App = dynamic(() => import('./App'), { ssr: true });
```
dadiorchen, line 13: I'm not sure this is the correct way, but it worked.

dadiorchen:
@cpaule1811 I will merge this PR first, we can improve it later, I am not sure my code is correct, but It works now, and can you add the layout.js based on this code? Actually I don't know how to do it.

src/Home.js
dadiorchen: I mean now the left side with text has been rendered on the server side!

cpaule1811:
Yeah that should be fine.

PR #142, opened 2021-09-23
feat: navbar component

cpaule1811:
Hi, I have made Navbar component. 

I added menu component from material ui as well just so it is mobile responsive but we can probably improve on this in future
Navbar
menu-mobile
.
menu-mobile-open


dadiorchen:
Right, this just what I need in the home page, :D https://github.com/Greenstand/treetracker-web-map-client/pull/141

PR #147, opened 2021-09-24
Issue11/logo enhancements

mmayeda:
- Map logo is framed
- Map logo component always displays "powered by greenstand" logo.  If there is also a "custom" logo loaded it displays that as well.

Screen Shot 2021-09-23 at 8 08 18 PM
Screen Shot 2021-09-23 at 8 09 48 PM

Screen Shot 2021-09-23 at 8 09 33 PM


dadiorchen:
@mmayeda super awosome!  @ZavenArra what do you think?

shreyaansjain06 (approved):

ZavenArra:
Looks great.  When it is the logo lockup in the upper right vs lower right corner?

mmayeda:
It is in the lower right for larger width screens and in the upper right for smaller ones (this is the same as current behavior).  The threshold should be 600px

PR #149, opened 2021-09-24
Cleanup code

dadiorchen:


dadiorchen:
@cpaule1811 do we need the `public` folder anymore after switching to next.js?

Issue #158, opened 2021-09-27
A flash glitch when open the home page of the built nextjs site.


dadiorchen:
**Describe the bug**
When open the built nextjs site, there is a flash blank page before the real home page loaded

**To Reproduce**
1. Build the project: npm run build
1. Run the next production server: npm start
1. Open localhost:3000

**Screenshots**

https://user-images.githubusercontent.com/5744708/134864350-62433dc9-1c36-47a3-8986-27589dce4219.mov

dadiorchen:
@cpaule1811 any idea?

PR #160, opened 2021-09-27
fix: #158 flash glitch

cpaule1811:
The problem was that material ui uses jss for styling so the styles were being loaded post javascript loaded. We just needed to add a custom document and inject the styles there so they can be rendered server side.


dadiorchen:
That's a lightning-speed PR!

PR #162, opened 2021-09-28
chore: moved background image out of layout to home component

cpaule1811:
This part of the home component was mistakenly in layout.js.


dadiorchen:
Yes, the background, thanks!

PR #167, opened 2021-09-29
resolves #165

### LINKED ISSUE #165: Can not visit trees/id directly
http://localhost:3000/trees/1

It's working to jump to this page by home page ---click--> top page ---click---> trees/1 , but can not open it directly in the browser, this is because in this way, the map context still isn't ready when try to manipulate the map object.

jamalvarez:
I might be wrong but I think that just by using `mapContext.map` and listening for changes we can do `flyTo` when it is available?

Just trying to try my luck with a PR (: 

Resolves #165

### DIFF
--- a/src/pages/trees/[treeid].js
+++ b/src/pages/trees/[treeid].js
@@ -10,8 +10,11 @@ const Tree = ({tree}) => {
 
   React.useEffect(() => {
     //manipulate the map
-    mapContext.map.flyTo(tree.lat, tree.lon, 16);
-  }, [] );
+    if(mapContext.map){
+      mapContext.map.flyTo(tree.lat, tree.lon, 16);
+    }
+
+  }, [mapContext.map] );
 
   return (
     <div style={{color:"white"}}>


dadiorchen:
That's a smart idea, did you try it on your computer, if it works, you should be able to open the URL: http://localhost:3000/trees/1 directly, did it work? 
I'm not sure that listening on non-prop-variable works for the useEffect?

jamalvarez:
I had tried and it was working, and just to make sure I did a small check again and this works as intended @dadiorchen:


image

image

when opening http://localhost:3000/trees/1

dadiorchen:
That's the smartest and best PR I found this month! 👍 

@jamalvarez are you interested to join our community? If true, please give me your email I want to invite you to our slack org.

dadiorchen:
But I still have a little confusion on it, how react hook useEffect be aware of a context value? I suppose we should add component props as the listening?

jamalvarez:
@dadiorchen it's because this component itself gets re-rendered, and then a run of the useEffect hook happens as it detects mapContext.map changed on these re-renders (:

So it seems there's an issue now regarding data because I try running the project locally and I'm getting this error:
image


because of rate limiting of this API:
`https://48b2db50-8226-4f1e-9b46-7d80bed46d0f.mock.pstmn.io/trees/1` 

image

So an issue needs to be created to tackle scenario when there is no response, or latitude and longitude are not present on the `tree` prop

dadiorchen:
We have noticed that I mean the API error and will solve it soon. 

In case you missed the message:
```
@jamalvarez are you interested to join our community? If true, please give me your email I want to invite you to our slack org.
```

jamalvarez:
Ah, sorry, didn't see that; marinalvjesus@gmail.com is my email

so if you want to invite me I'm flattered even if I'm not sure of how much help I could be (:

Issue #154, opened 2021-09-27
Badge component.


dadiorchen:
Please put this component to `common` folder for future reuse.
This component should have status: enabled, the picture below shows the status: enabled=true, when enabled=false, the component should be in gray color indicating that the tree doesn't have this Badge yet.
<img width="1440" alt="Screen Shot 2021-09-27 at 10 58 26 AM" src="https://user-images.githubusercontent.com/5744708/134839348-8e51eb23-ac0e-4000-88dc-fcb7e3b03021.png">

---
Some hints for building components:
* Design resource: https://www.figma.com/file/XdYFdjlsHvxehlrkPVYq0l/Greenstand-Webmap?node-id=2497%3A11584
* A tutorial of how to use the Cypress component tool to build components: 
  https://github.com/Greenstand/treetracker-wallet-web#how-to-build-a-component
* We are using MaterialUI to build the app: https://material-ui.com/
* When you are ready to raise PR for the component, please attach the screenshot of the component for review.
* Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
* Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
* To know more about our organization, visit our [website](https://greenstand.org).

PR #171, opened 2021-09-30
feat: badge component #154

cpaule1811:
Tree Verified

Added Badge component. The reason the photo is the wrong color is because The ThemeProvider isn't mounted by Cypress, Some of the other component tests that were dependent on the theme had the same issues so we probably need to raise this as an issue. If you put the component in the applciaion it will be green when enabled and low opacity white/grey when disabled.


dadiorchen:
@cpaule1811 make totally sense, I will open that issue.

dadiorchen:
By they way, how do you run the component tool ?

PR #172, opened 2021-09-30
fix: remove zoom control panel #93

jwgatt:
Fixes [#93](https://github.com/Greenstand/treetracker-wallet-web/issues/93)

### DIFF
--- a/src/models/Map.js
+++ b/src/models/Map.js
@@ -109,13 +109,8 @@ export default class Map {
       center: this.initialCenter,
       zoomControl: false,
     };
+    
     this.map = this.L.map(domElement, mapOptions);
-
-    // control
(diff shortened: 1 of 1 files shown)


dadiorchen:
@jwgatt I suppose this will change the behavior on the current repo? So there isn't zoom button on the map by default? 
If is true, I think it will impact the current treetracker.org, right?

jwgatt:
@dadiorchen The zoom control appears in the top left hand control by default. We can the add `zoomControl: false` to disable it, in this block of code found in Map.js;
 ```  async mount(domElement) {
    const mapOptions = {
      minZoom: this.minZoom,
      center: this.initialCenter,
      zoomControl: false,
    };

dadiorchen:
Beautiful!

PR #183, opened 2021-10-09
Link component

### LINKED ISSUE #181: Link passing Href to a tag
Next Link component won't always generate html a tag as child. This is needed for web crawlers to pick up the website for better SEO. 

The current code

<Link href="/" passHref>
         <Button className={classes.buttonStyle} variant="text" component="a">
            Contact Us
        </Button>
</Link>

without a tag

This code fixes it by using component props with any Mui component that is wrapped in a Next link component

<Link href="/" passHref>
         <Button className={classes.buttonStyle} variant="text" component="a">
            Contact Us
        </Button>
</Link>

with a tag

@QuaidBartolomei mentioned another potential solution where we use custom Link components from Mui which I think by default is an anchor tag so that should work to I believe.

QuaidBartolomei:
resolves #181

### DIFF
--- a/src/components/Home.js
+++ b/src/components/Home.js
@@ -1,12 +1,12 @@
-import React from 'react';
-import { makeStyles } from '@material-ui/core/styles';
-import Grid from '@material-ui/core/Grid';
-import Typography from '@material-ui/core/Typography';
-import Button from '@material-ui/core/Button';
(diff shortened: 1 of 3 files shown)


dadiorchen:
@cpaule1811 take a look at this.

PR #188, opened 2021-10-10
organizations/[id] page

cpaule1811:
Do we have a mock api for organization? Just waiting on that and a couple of components to complete this page. 

I set the Mui's typography variant 'sub headers' in the theme seeing as this typography shows up in not just this page but other places in the application. I think we could probably use this to style all the necessary variants in the theme to be honest that way we won't have to style typography for each component and page, what are your thoughts?


cpaule1811:
organizations draft

dadiorchen:
@cpaule1811 yes let's try this global styling, thanks.

PR #190, opened 2021-10-10
#resolves#152#cluster highlight issue

danp2392:
Fixed the cluster highlight issue by adding the correct css code and icons

image


dadiorchen:
@danp2392 thank you very much, good job!

PR #197, opened 2021-10-13
fix: added .prettierignore

QuaidBartolomei:
This fixes the issue of commits taking a long time to resolve.


QuaidBartolomei:
I'm changing this to a draft because there are more steps I can take to improve this issue

PR #200, opened 2021-10-14
fix: updated husky and added lint-staged

QuaidBartolomei:
I updated husky and configured it with lint-staged so that only files with staged changes will be processed by eslint and prettier when using `git commit`. This will make commits much faster especially on slower hardware.


dadiorchen:
@QuaidBartolomei can you help to resolve the conflict in this merge? It's not your fault, it just emerges when I merged some other branch prior this PR

QuaidBartolomei:
@dadiorchen It's all good now

PR #221, opened 2021-10-22
fix: full path in pushState when URL updates

monjurkhuda:
Resolves #218

Simply commented out the  updateUrl( ) function.


dadiorchen:
@monjurkhuda great! thank you, I will merge this PR!

Issue #229, opened 2021-10-25
Cypress will not run new component test files (no tests found)


QuaidBartolomei:
Currently if I try to make a new component test file (or duplicate an existing one) Cypress component tests (`cyu`) will say "No tests found" for that file.

If I delete my .next/ folder and restart cypress-ct the tests will work like normal but this must be done any time a new test file is created.

This seems to be an issue with the Cypress package and a fix is awaiting merge: https://github.com/cypress-io/cypress/issues/17920

dadiorchen:
@QuaidBartolomei can we close this one now?

QuaidBartolomei:
@dadiorchen Not yet, the issue of Cypress not working as intended still exists and we have `npm run cyux` as a workaround. We can close this after the cypress issue is patched.

PR #231, opened 2021-10-25
fix: cypress-ct font loading

### LINKED ISSUE #230: Cypress component testing not rendering correct font
The web app is now using the correct font (Lato) but cypress-ct is not loading it. An easy way to test this is to change this line in `theme.js` so the backup font will be serif:

```
  typography: {
    fontFamily: ['Lato', 'Serif'].join(','),
```

QuaidBartolomei:
resolves #230 

I resolved this by adding webfontloader as a dev package and using the cypress support file to run it before each test.

I also added `"cyux": "rm -rf .next/; npm run cyu",` as a helper script until #229 is resolved

### DIFF
--- a/cypress/support/components-testing.js
+++ b/cypress/support/components-testing.js
@@ -16,3 +16,11 @@
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
 import './commands';
+
+import WebFont from 'webfontloader';
+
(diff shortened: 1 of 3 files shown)


dadiorchen:
@QuaidBartolomei cool

Issue #123, opened 2021-09-17
The /top page.


dadiorchen:
* Prerequisites: #103 #121 
* To integrate the /top page, using mock API service.
* Write Cypress integration test cover this page. ( check the readme for the integration test guide)
* Check API document for API spec.
* Design resource: https://www.figma.com/file/XdYFdjlsHvxehlrkPVYq0l/Greenstand-Webmap?node-id=2497%3A11584

<img width="1440" alt="Screen Shot 2021-10-19 at 4 57 48 PM" src="https://user-images.githubusercontent.com/5744708/137877636-2b6cc5c1-ca06-4457-a9f9-fa1908b401b8.png">

---
Some hints:
* Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
* Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
* To know more about our organization, visit our [website](https://greenstand.org).

echai-11:
Is anyone working on this? Otherwise I would like to work on it.

dadiorchen:
Hi, @echai-11 this is a valid task to take, please.
And I have done some basic work for this page, please update the newest code from `web-map-site`
And I guess @jwgatt has finished the leader board component.

echai-11:
Noted, then I will work on this task.

dadiorchen:
@echai-11 do you get any progress on this? Any help do you need?

dadiorchen:
This issue now is open for everyone!

dadiorchen:
#299  solved this issue

PR #252, opened 2021-11-04
Top page

dadiorchen:
Related to #123 , but didn't finish it.


dadiorchen:
@QuaidBartolomei please review this code, but I have to go forward so I will merge it first, we can discuss and change the code later.

Issue #255, opened 2021-11-04
Some problem when auto-refreshing the page by next.js dev mode.


dadiorchen:
When modifying js file, next.js will automatically reresh the web page, but sometimes it show error:
![image](https://user-images.githubusercontent.com/5744708/140249439-83c3bcb4-f26d-463d-9a92-00cb00e10884.png)

This is related to the map object mounting to the DOM, maybe we need a better way to mount the map object to the DOM

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.

LahiaOmar:
Hey @dadiorchen, I would like to start working/fixing this bug

PR #262, opened 2021-11-04
fix: auto-refreshing the page in dev mode #255

LahiaOmar:
Fix the bug, by testing if the  MapContext has already updated the map object.
I think it's can be fixed in a better way, by moving the creation of the Map object inside MapContextProvider, and providing for the consumer's context just a init function instead of setMap, but NextJS don't allow the import of the class Map.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -365,6 +365,7 @@ function MapComponent() {
 
   // load map
   React.useEffect(() => {
+    if (mapContext.map) return;
     log.info('load map...');
     // disable waiting for loading
     loaded();


jwgatt (approved):

dadiorchen:
Great!

dadiorchen:
@LahiaOmar what's your name in slack?

LahiaOmar:
He @dadiorchen  it's Omar Lahia

dadiorchen:
Hi, misunderstanding :D I mean have you joined our slack workspace? Greenstand.

LahiaOmar:
HHH it's me :p, where can I join?

Issue #261, opened 2021-11-04
Error thrown on clicking tree icon 


jwgatt:
I'm getting this error on clicking the tree icon in dev mode running the mock server.

It might be due to this;
> Note: _document.js is only rendered on the server side and not on the client side. so event handlers like onClick is not going to work.
https://stackoverflow.com/questions/51040669/next-js-use-of-app-js-and-document-js

![Screenshot from 2021-11-04 10-45-13](https://user-images.githubusercontent.com/79326421/140292375-d7d0a6a3-2247-4b1a-bbe1-ec0dcb336e1a.png)

echai-11:
I am working on this bug.

PR #265, opened 2021-11-07
fix: #261 error on clicking tree icon

echai-11:
The error that I see is that mock server does not have data to display to the tree information. So, I just added checks to make sure that data exists and then to render image, otherwise image will not render. 
Screen Shot 2021-11-07 at 1 28 40 PM


jwgatt (approved):

dadiorchen:
thanks @echai-11 are you in the slack? what's your name over there?

echai-11:
@dadiorchen my name on slack is Lizzie Chai

Issue #253, opened 2021-11-04
Might be a next.js wrongdoing.


dadiorchen:
When the first time to visit the home page, and click the `find tree` button, the page will get refreshed rather then go forward to the link.

https://user-images.githubusercontent.com/5744708/140245554-83c68f81-b062-44f3-81cc-e601ac1b31ef.mov

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.

echai-11:
I would like to work on this task.

QuaidBartolomei:
This issue was fixed by the new setup with mui5 from this PR #287

PR #272, opened 2021-11-09
fix: #253 on initial page load clicking on link rerenders page

echai-11:
The issue is related to getInitialProps. From next.js doc:

"For the initial page load, getInitialProps will run on the server only. getInitialProps will then run on the client when navigating to a different route via the next/link component or by using next/router. However, if getInitialProps is used in a custom _app.js, and the page being navigated to implements getServerSideProps, then getInitialProps will run on the server."  (https://nextjs.org/docs/api-reference/data-fetching/getInitialProps)

 This is the log from what I saw:
Screen Shot 2021-11-09 at 11 04 43 AM

Not sure if this is ideal fix... But this is the problem I saw.

### DIFF
--- a/src/pages/_document.js
+++ b/src/pages/_document.js
@@ -25,23 +25,3 @@ export default class MyDocument extends Document {
     );
   }
 }
-
-MyDocument.getInitialProps = async (ctx) => {
-  const sheets = new ServerStyleSheets();
-  const originalRenderPage = ctx.renderPage;
-
-  ctx.renderPage = () =>
-    originalRenderPage({
-      enhanceApp: (App) => (props) => sheets.collect(<App {...props} />),
-    });
-
-  const initialProps = await Document.getInitialProps(ctx);
-
-  return {
-    ...initialProps,
-    styles: [
-      ...React.Children.toArray(initialProps.styles),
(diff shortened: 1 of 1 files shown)


jwgatt (approved):

QuaidBartolomei:
This type of setup is used in [material-ui's nextjs examples](https://github.com/mui-org/material-ui/blob/5f59a29923bf7755d2d5e9c8d605fd15e924c9da/examples/nextjs/pages/_document.js). These changes may affect how styles are applied to pages using SSR.

echai-11:
> This type of setup is used in [material-ui's nextjs examples](https://github.com/mui-org/material-ui/blob/5f59a29923bf7755d2d5e9c8d605fd15e924c9da/examples/nextjs/pages/_document.js). These changes may affect how styles are applied to pages using SSR.

Yes, as seen in the log, I think getInitialProp is preventing getServerSideProps (fetching data from server) from being fetched on top page on the first time it is being called. On first click, top page is compiled but data is not fetched. Only on second click, data is being fetched on top page after getInitialProp is finished running.

QuaidBartolomei:
This issue was fixed when the project migrated to mui5 in this PR #287

Issue #275, opened 2021-11-10
Write Jest test to cover requestAPI function


dadiorchen:
https://github.com/Greenstand/treetracker-web-map-client/blob/4e6f680d124ae6d74c92bd501eda627919b32f26/src/models/utils.js#L46-L62

Write test in utils.test.js to cover the `requestAPI` function. Please note, there is a bug in the function, if the the request failed, like, receiving 404, for example, the server response with: code:404, json body: {msg: "some error message"}, the function will still successfully respond with the json body, this is a wrong behavior, the correct behavior should be: throw an error from this function.

It would be great if we can also write test code to cover this bug.

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.

LahiaOmar:
Hi @dadiorchen, I will try to work on this issue.

dadiorchen:
Finished

PR #286, opened 2021-11-18
fix: add test to requestAPI, and modify requestAPI to use axios … #275

LahiaOmar:
- To test this function, the use of fetch in jest/nodeJS environment will not work. fetch don't exist. instead, if we use Axios, it's throw an error when the server returns a 4XX/5XX response ( for the 404 bugs).
- for the test a just test it with a wrong URL, tell me if they are other URLs to test I will make changes for that.

### DIFF
--- a/src/models/utils.js
+++ b/src/models/utils.js
@@ -1,4 +1,5 @@
 import { catchClause } from '@babel/types';
+import axios from 'axios';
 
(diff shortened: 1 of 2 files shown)


dadiorchen:
@LahiaOmar great job!

Issue #236, opened 2021-10-28
Improve the location component


dadiorchen:
We need to improve the location component to passing in coordinates rather than a string, and parse the coordinates then translate to the location name.

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.

dadiorchen:
@ojafero I will close this issue, we are going to rebuild this feature.

PR #323, opened 2021-12-01
Fix: Improve location component 236

ojafero:
fixes: #236 
location-fix


dadiorchen:
@ojafero you are the first guy trying to update the yaml, and mock server :D thanks for your contribution.
I rethought the API here, I'm afraid it's a bit hard for us to locate the city of coodirates, so I need you to make some change in this PR:

Remove the /locations/xxx API, then use the /coutries?lat=xx&lon=xxx API, so we can get the country name of the location, and ignore the city name.

PR #343, opened 2021-12-04
mobile drawer

dadiorchen:
This is for the mobile layout with the drawer.

https://dadior.s3.ap-northeast-1.amazonaws.com/Screen%20Recording%202021-12-04%20at%204.47.24%20PM.mp4

Some code is for temporarily presenting. And the whole solution could be replaced with a better approach in the future.


src/pages/index.js
```diff
@@ -25,3 +25,5 @@ export default function Homepage() {
     </>
   );
 }
+
+Homepage.isBLayout = true;
```
dadiorchen, line 29: An ugly approach

.eslintignore
```diff
@@ -2,4 +2,5 @@
 dist/
 coverage/
 node_modules/
-package-lock.json
\ No newline at end of file
+package-lock.json
+.eslintrc.json
```
QuaidBartolomei, line 6: This has not been an issue for me in my other projects. I think the Typescript eslint parser I normally use excludes .json files but this project is Javascript and therefore uses a different parser so eslint will try to scan them if they are passed on the command line which lint-staged will do.

QuaidBartolomei (changes requested):
2021-12-05 00 13 03 localhost d4799b185e2c

Something caused the homepage layout to change.

dadiorchen:
> 2021-12-05 00 13 03 localhost d4799b185e2c
> 
> Something caused the homepage layout to change.

thanks! fixed:
image

src/components/Home.js
```diff
@@ -85,6 +86,7 @@ export default function Home() {
           Come explore the global reforestation effort.
         </Typography>
         <Buttons />
+        <Box sx={{ height: 1000 }} />
```
QuaidBartolomei, line 89:
```suggestion
```

QuaidBartolomei: remove this and `height: '100%'` @ line 19 to fix the layout issue

QuaidBartolomei (changes requested):
2021-12-05 19 37 18 localhost bb7603faefd8
Now there's a different layout issue

VWRoli (approved):

PR #348, opened 2021-12-05
Added flag icons in .svg format

Koda-Pig:
Downloaded all flags from https://flagicons.lipis.dev/ and added them to the src/pages/flags directory.
Included 1x1 and 4x3 aspect ratios.
Renamed all flags from their 2 letter code to the full name.

Fixes this issue: https://github.com/Greenstand/treetracker-web-map-client/issues/298

PR #351, opened 2021-12-06
Moved flags to public directory

Koda-Pig:
Moved flags from src to public directory so that the flags are accessible on the front end. Eg: http://localhost:3000/images/flags/4x3/china.svg


QuaidBartolomei (approved):

QuaidBartolomei (changes requested):
[This comment](https://github.com/Greenstand/treetracker-web-map-client/pull/348#issuecomment-986557745) from #348 should be addressed.

Issue #311, opened 2021-11-26
A reusable component for image


dadiorchen:
We need a reusable component to apply to any image element in the system, the function is that:
* A taken on icon indicating when do we take this photo
* A maximum icon that we can click to open a window displaying the full size of the picture.
* A like icon that we need the work from #310 

<img width="1440" alt="Screen Shot 2021-11-26 at 5 19 55 PM" src="https://user-images.githubusercontent.com/5744708/143557777-84b49825-2f6e-4f93-9bd9-de5593004638.png">
<img width="1440" alt="Screen Shot 2021-11-26 at 5 20 10 PM" src="https://user-images.githubusercontent.com/5744708/143557781-300d38c2-5348-44b4-959d-139dfc77fefa.png">

---

Some hints for building components:

- This kind of issue is about creating a new component, it might not need to mount it in the whole app, to know how to create a component separately, please check the tutorial below.
- Design resource: https://www.figma.com/file/XdYFdjlsHvxehlrkPVYq0l/Greenstand-Webmap?node-id=2497%3A11584
- A tutorial of how to use the Cypress component tool to build components:
  https://github.com/Greenstand/treetracker-wallet-web#how-to-build-a-component
- We are using MaterialUI to build the app: https://material-ui.com/
- Please mount the component with `import { mountWithTheme as mount } from '../../models/test-utils';` to get our theme imported.
- Please use MaterialUI `theme.spacing(x)` as possible as we can, so the component is more responsive
- When you are ready to raise PR for the component, please attach the screenshot of the component for review.
- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.
- More issues in this project: https://github.com/orgs/Greenstand/projects/26

pierrelstan:
I would like to work on this issue, my email to join the slack community: stanleypl75@gmail.com

dadiorchen:
@pierrelstan thank you, it's yours

QuaidBartolomei:
@dadiorchen can this issue be closed?

dadiorchen:
Sure!

Issue #340, opened 2021-12-03
Implement the full-functional filter component, with unit tests. 


dadiorchen:
Please check the video describing what needs to be done, basically, some interaction of the component, some verification of the filter conditions, some text job. 

Also, please cover the feature with the component test in Filter.cy.js

https://dadior.s3.ap-northeast-1.amazonaws.com/Screen%20Recording%202021-12-03%20at%204.40.40%20PM.mp4

![image](https://user-images.githubusercontent.com/5744708/144573199-a88fc251-6ce7-4c13-90c4-1b410eee4820.png)

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.
- More issues in this project: https://github.com/orgs/Greenstand/projects/26

PR #358, opened 2021-12-10
Reusable Image component 

pierrelstan:
# Description
Create a reusable image component that has:
- An icon indicating when do we take this photo
- A maximum icon that we can click to open a window displaying the full size of the picture.
- A like icon 

Fixes # (#311)

# How Has This Been Tested?

- [ ] Cypress integration
- [ ] Cypress component tests

# Checklist:

- [ ] My code follows the style guidelines of this project


dadiorchen:
@pierrelstan please solve the conflit in this PR ( in case you are not familiar with this), you can pull the web-map-site branch and merge/rebase from it, and solve the conflict locally then push this branch again)

Please attach a screenshot of your component

VWRoli (changes requested):
Naming: it is a reusable image component, not a tree image component. The naming should reflect that. `CustomImageWrapper` or something.

In the commit message reference the issue number so it links it to that issue (or in the PR request? Not sure. Best if it's in both places).

Merge the web-map-site branch into yours as the theme got updated. (Typography and colors)

No need for the template literals here:
https://github.com/Greenstand/treetracker-web-map-client/blob/24f88ca15859f7adb3aa2e51639348427979a44b/src/components/common/TreeImage.js#L34

___

For texts use Typography, from MUI, not divs. With Typography, you can easily style the element, without adding classes. For example:
https://github.com/Greenstand/treetracker-web-map-client/blob/24f88ca15859f7adb3aa2e51639348427979a44b/src/components/common/TreeImage.js#L193
Could be  `<Typography variant="caption">Photo taken on</Typography>`

___

https://github.com/Greenstand/treetracker-web-map-client/blob/24f88ca15859f7adb3aa2e51639348427979a44b/src/components/common/TreeImage.js#L194-L196
Could be: `<Typography variant="h6">
              {dayOfWeek}/{month}/{year}
            </Typography>`

___

Here, Grid has its highly customizable properties. Add the alignments in line, also add the container property, and its child should be `<Grid item></Grid>` and not divs.
https://github.com/Greenstand/treetracker-web-map-client/blob/24f88ca15859f7adb3aa2e51639348427979a44b/src/components/common/TreeImage.js#L188

___

Don't hardcode the colors. Especially if you have to use it in three different places. If it is not in the theme put it somewhere where it is most logical. In this case, it's already in the theme `textPrimary`. The theme got updated from when you started this branch.

___

https://github.com/Greenstand/treetracker-web-map-client/blob/24f88ca15859f7adb3aa2e51639348427979a44b/src/components/common/TreeImage.js#L119

`wrapperFavorite` , `containerOpenWithOutlinedIcon`, `ContainerIcon` has a lot of common styles, which could be a common class, so we don't violate the DRY principle.
___

Don't bother with letterspacing, as I noticed in the Figma design letter spacing is quite universal, I will take a look at the theme, and add letterspacing globally, so in the future, we will not have to bother with it.

Margins and padding should be `theme.spacing()`, I'm guessing positioning items too.

___

Generally speaking good, job. The most important thing is to use MUI with all its power and customizability. You can get rid of a lot of hardcoded styles and classes by using MUI, which makes the code more readable.

pierrelstan:
Screen Shot 2021-12-11 at 5 00 18 PM

VWRoli (changes requested):
You can add a Typography element here and you can remove the class of the likes.
https://github.com/Greenstand/treetracker-web-map-client/blob/5fff591acb2aa14d3d22e01f91a6270655b8d0cf/src/components/common/CustomImageWrapper.js#L174-L176

Like this: `<Typography color="white">{likes}</Typography>`

___

On lines 86 and 95 you should use `theme.palette.common.white` instead of hardcoded color values.
https://github.com/Greenstand/treetracker-web-map-client/blob/5fff591acb2aa14d3d22e01f91a6270655b8d0cf/src/components/common/CustomImageWrapper.js#L86

Same goes for line 73:
https://github.com/Greenstand/treetracker-web-map-client/blob/5fff591acb2aa14d3d22e01f91a6270655b8d0cf/src/components/common/CustomImageWrapper.js#L73
This color is `theme.palette.textPrimary.main` with the opacity of 0.6
Again use the colors from the theme.

___

Also, I just realized that we can further improve the code. 
The wrappers/containers for the `likes`, `icon`, and `date` can be one common component. A component that accepts 4 positional (all of them optional) props. 

You can put it in the same file as we will only use it here. They are doing basically the same thing just in a different position.

And instead of having fixed widths, you can decrease the font size a little bit and adjust the padding/margin on small screens so it's more responsive.

Something like:
`<InfoWrapper top={1} left={1} right={1} bottom={1}></InfoWrapper>` 
(If you have a better name for it go for it)
You can position them like top={x} right={x}, and don't have to give them the left and bottom. (I think it's doable.)

Great job with the improvements by the way! Keep up the good work! Let me know if you have any questions.

pierrelstan:
@VWRoli  thanks for the feedback.

dadiorchen:
@VWRoli thank you for your great review and explanation! I also learned a lot!

VWRoli (approved):
Great job! Thanks for the improvements!

pierrelstan:
> Great job! Thanks for the improvements!

The reviews and explanations helped me to get to that improvements.
 Thank you for leading me to do it the right way.

dadiorchen:
Hey, @pierrelstan I just noticed that you are really using the cypress to build the test, that's a super good way to lead us to a higher standard! Thank you!

Just in case you are interested, this issue is about using cypress to build and check the logical of the filter component, the logic in this component would be more complicated. : ) https://github.com/Greenstand/treetracker-web-map-client/issues/340

dadiorchen:
Just some suggestions/ideas about how to check the result of the test, I think the better way is to check some element that got semantic, rather than technical elements. 
A example for the former: planter avatar
A example for the later: 

The reason is that: the former is more worth checking, and the former would be more stable, so your test has higher possibility to survive along the way of iterations.

I remember there are some discussions in the testing-library document: https://testing-library.com/docs/queries/about/#textmatch

pierrelstan:
> Hey, @pierrelstan I just noticed that you are really using the cypress to build the test, that's a super good way to lead us to a higher standard! Thank you!
> 
> Just in case you are interested, this issue is about using cypress to build and check the logical of the filter component, the logic in this component would be more complicated. : ) #340

@dadiorchen I'm interested in doing it. Assign it to me.

PR #387, opened 2021-12-28
Hotfix: Make workflow tests pass

### LINKED ISSUE #386: CustomImageWrapper component test fails
The newly added workflow tests are not passing. It is important that we get these tests to pass and keep it that way for the workflows to be effective. Here is the failed workflow action from a [test PR](https://github.com/QuaidBartolomei/treetracker-web-map-client/pull/5) on my fork https://github.com/QuaidBartolomei/treetracker-web-map-client/actions/runs/1629151642 . It seems to be an issue with loading the test image for `CustomImageWrapper`. When I run the tests locally the image does not load, but if I wait a while it eventually does load and the test passes. Currently I am testing if adding a longer timeout to this test will work, if not I will mark it as skippable until it is fixed.

QuaidBartolomei:
# Description

Fix failing tests for PR workflow. The CustomImageWrapper test needed a longer timeout for the image to load.

See issue for full description.

Fixes #386 

## Type of change


src/components/common/CustomImageWrapper.cy.js
```diff
@@ -22,21 +22,27 @@ describe('CustomImageWrapper', () => {
...
-      .should('be.visible')
-      .and(($img) => {
-        expect($img[0].naturalWidth).to.be.greaterThan(0);
-      });
-  });
+  it(
+    'Image renders properly',
+    {
+      defaultCommandTimeout: 180000,
+    },
```
QuaidBartolomei, line 29: this applies a special config to this test only. I tried 1 minute and the test still failed locally so it's 3 minutes for now.

PR #399, opened 2022-01-11
Update eslint and eslint rule settings, Update NextJS to v12, sort package.json

QuaidBartolomei:
# Description

- updated eslint and the airbnb presets.
- replaced eslint `simple-import-sort` with configurable `import/order` rule
- changed the eslint config to be simpler and use recommended plugin configs.
- re-added `--cache` to the lint-staged settings, The issue we were having previously was from the `--quiet` option causing auto fixable warnings to be ignored.
- cleaned up npm scripts in package.json
- removed unused dependencies
- updated nextjs to v12 for faster builds. There should be no issues with updating https://nextjs.org/docs/upgrading#upgrading-from-11-to-12

The remaining eslint warnings are all issues that should be fixed.

One of which is replacing `` elements with the `<Image>` component from `next/image`
https://nextjs.org/docs/messages/no-img-element

Here are the main reasons that from the docs that we should use `next/image`:

> - Improved Performance: Always serve correctly sized image for each device, using modern image formats.
> - Visual Stability: Prevent Cumulative Layout Shift automatically.
> - Faster Page Loads: Images are only loaded when they enter the viewport, with optional blur-up placeholders
> - Asset Flexibility: On-demand image resizing, even for images stored on remote servers

https://nextjs.org/docs/basic-features/image-optimization

Here is the api reference.
https://nextjs.org/docs/api-reference/next/image


dadiorchen:
Thanks @QuaidBartolomei this is a big improvement

Siddhant-K-code (commented):
LGTM 🚀 

Thanks ✨

PR #436, opened 2022-01-22
solved issue 434

Krish-bhardwaj:
# Description
changed the letter spacing to 0

solved issue #434

Issue #434, opened 2022-01-22
Need to remove this default CSS property


dadiorchen:
The typography element will automatically add the letter-spacing, but we can not find it in the project, we need to set it to '0';

<img width="1440" alt="Screen Shot 2022-01-22 at 11 04 39 AM" src="https://user-images.githubusercontent.com/5744708/150622555-38a27e0d-b967-4404-9af2-3d851419d1d8.png">

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.
- More issues in this project: https://github.com/orgs/Greenstand/projects/26

Krish-bhardwaj:
Hi i am a fresher looking for some opensource contribution can u help me out i want to contribute to your project and can u assign me this project :)

dadiorchen:
:) @Krish-bhardwaj sure, thank you!

codyconway77:
I found a fix. Don't want to step on any toes and take away the opportunity from Krish-bhardwaj. Would it be ok to open a PR with my fix in just in case the issue goes stale?

Krish-bhardwaj:
@dadiorchen i have solved this issue ( i am in my very early stages of my contribution this is my second opensource contribution requested to merge my PR )

codyconway77:
I believe this is the fix. Referenced the MUI docs at https://mui.com/customization/theme-components/#global-style-overrides
```
const theme = createTheme({
  components: {
    // Name of the component
    MuiButton: {
      styleOverrides: {
        // Name of the slot
        root: {
          // Some CSS
          fontSize: '1rem',
        },
      },
    },
  },
```

codyconway77:
#435

dadiorchen:
@dark-teal-coder thanks, it's intresting, I wonder where the original left-spacing is from?

keyy123:
Is this issue still open? I don't mind giving it a try! 😁👌

dadiorchen:
@keyy123 thank you for helping on this

PR #437, opened 2022-01-22
solved issue #436

Krish-bhardwaj:
# Description
changed the letter spacing to 0 solved issue #434

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -43,8 +43,9 @@ const colorTheme = createTheme({
 const theme = createTheme(colorTheme, {
   typography: {
     fontFamily: ['Lato', 'Roboto', 'Helvetica', 'Arial', 'sans-serif'].join(
-      ',',letterSpacing: 0,
+      ','
     ),
+    letterSpacing: 0,
     fontWeight: 400,
     h1: {
       fontFamily: ['Montserrat'].join(','),
--- a/src/theme.js
+++ b/src/theme.js
@@ -43,7 +43,7 @@ const colorTheme = createTheme({
 const theme = createTheme(colorTheme, {
   typography: {
     fontFamily: ['Lato', 'Roboto', 'Helvetica', 'Arial', 'sans-serif'].join(
-      ',',
+      ',',letterSpacing: 0,
     ),
     fontWeight: 400,
     h1: {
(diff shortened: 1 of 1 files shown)


Krish-bhardwaj:
hi @QuaidBartolomei  i have made the suggested changed plz look after my PR

dadiorchen:
@Krish-bhardwaj no need to close and create PR next time, just directly push up your branch works

QuaidBartolomei:
@Krish-bhardwaj I'm still encountering the same issue with these changes.

2022-01-22 01 12 54 localhost 81644b97438f

dadiorchen:
@QuaidBartolomei yea, that's what surprised me, and I can not find the source of this settings

Krish-bhardwaj:
Hi i have made some change @dadiorchen and @QuaidBartolomei  ig this solves our issue !

src/theme.js
```diff
@@ -39,11 +39,24 @@ const colorTheme = createTheme({
...
+      styleOverrides: {
+        // Name of the slot
+        root: {
+          // Some CSS
+          letterSpacing: 0,
+        },
+      },
+    },
+  },
+});
```
dadiorchen, line 55: is it possible to declare two const with the same name?

VWRoli (changes requested):
https://github.com/Greenstand/treetracker-web-map-client/blob/3eeb17cdd55123a593c932ed531f36de7e8db2e6/src/theme.js#L42-L55

We can't create a new theme because we already have one. And also you try to change letterspacing on the button component, and the problem is the h1 component.

https://github.com/Greenstand/treetracker-web-map-client/blob/3eeb17cdd55123a593c932ed531f36de7e8db2e6/src/theme.js#L111
This also has nothing to do with the issue as this is for the body1 component, which should have a default letter spacing.

https://github.com/Greenstand/treetracker-web-map-client/blob/3eeb17cdd55123a593c932ed531f36de7e8db2e6/src/theme.js#L62-L70
H1 element probably has a default letterspacing defined by MUI, so we can't remove it, just override it.
If we put ` letterSpacing: 0` into this part of the code, where we define the default for `h1` component we override the default negative letter spacing.

Screenshot from 2022-01-29 14-37-45

src/theme.js
VWRoli: No, it is not.

PR #439, opened 2022-01-22
Added function to hide fullname

Niklas-dev:
I have added a function to hide the fullname of the planter. The function is taking in the fullname and returns the firstname and first letter of the lastname.

The edit is corresponding to the issue "Hide the full name of the planter"

- New feature (non-breaking change which adds functionality)


dadiorchen (changes requested):
@Niklas-dev I need you to do two extra things:
1. Use unit test to cover it, like: it('Dadior Chen should return Dadir C')
2. Apply this function to the planter page: http://localhost:3000/planters/940

Niklas-dev:
Alright, will do it.

Niklas-dev:
> 

This includes all the requested changes

VWRoli:
The function could be simplified like this.
const hideFullName = (name) => `${name.charAt(0)}.`;

And I would also add a dot at the end so we know it is shortened.

Also if you create a PR please reference the issue, so others do not create a PR for the same issue. Like myself:D

VWRoli (changes requested):
The function could be simplified like this.
const hideFullName = (name) => ${name.charAt(0)}.;

And I would also add a dot at the end so we know it is shortened. What do you think @dadiorchen?

Also if you create a PR please reference the issue, so others do not create a PR for the same issue. Like myself:D

Niklas-dev:
Okay thanks for the feedback. Was my first PR.

dadiorchen:
oh, @VWRoli you are right, I forgot the dot, thanks, guys.!

Niklas-dev:
> Member

I have tried it but that way the first name is also hiding so its still more complex.

VWRoli:
Take a look at this PR #447 I made, if something does not work you might find some good information there.

Niklas-dev:
> Take a look at this PR #447 I made, if something does not work you might find some good information there.

Ah, now i got what you meant i thought about passing the full name to the function but you want just to pass the lastname.

jwgatt (approved):

dadiorchen:
@Niklas-dev thanks, great job!

PR #462, opened 2022-02-07
Tests mock api calls with msw

QuaidBartolomei:
# Description

- replace `nock` with `msw`
  - the msw server will intercept any external requests that match its defined route handlers in `src/mocks/handlers.js`
  - Nextjs will start the mock server in `pages/_app.js` if env var `NEXT_PUBLIC_API_MOCKING === 'enabled'`
  - Cypress tests will not need to use nock or implement their own server mocks because the nextjs server will handle them
  - Jest tests can manually initialize the mock server by importing it from `src/mocks/server.js`
    - see `api.spec.js` for an example

## Type of change

## Screenshots

# How Has This Been Tested?

- [ ] Cypress integration
- [ ] Cypress component tests
- [ ] Jest unit tests

# Checklist:

### DIFF
--- a/.env
+++ b/.env
@@ -2,4 +2,3 @@ NEXT_PUBLIC_API=https://dev-k8s.treetracker.org/webmap/
 NEXT_PUBLIC_API_NEW=http://127.0.0.1:4010/mock
 NEXT_PUBLIC_TILE_SERVER_URL=https://{s}.treetracker.org/tiles/new/
 NEXT_PUBLIC_TILE_SERVER_SUBDOMAINS=dev-k8s
-REACT_APP_API_NEW=https://48b2db50-8226-4f1e-9b46-7d80bed46d0f.mock.pstmn.io/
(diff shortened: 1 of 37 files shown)


dadiorchen:
What's the advantage of msw comparing to nock?

QuaidBartolomei:
@dadiorchen The main advantage is that the mock server only needs to be written once, instead of in each individual cypress test as it is currently. Writing cypress tests will be much easier now. This also lets us use the same mock server for both jest and cypress testing. The goal is to make our test writing easier and faster and also simplify the server mocking implementation and configuration.

dadiorchen:
@QuaidBartolomei I'm correct, the cypress tests can share the same mock service by msw?
And is there a way to change the mock behavior for some special cypress tests? I mean, We can respond with some special mock data for a single test case.

QuaidBartolomei:
@dadiorchen unfortunately you are correct. This was something I personally just wanted to experiment with. I'm replacing this PR with #470 which will refactor and fix the cypress test mocks with `nock`.

Issue #425, opened 2022-01-20
Refine the planter photo layout


dadiorchen:
Now, it's:
* Too big
* Maybe rounded convert is better, please refer to the UI design in Figma.
* Maybe need to tweak image/background layout to fit into the photo frame, the original photo direction usually is 'portrait', but our design layout is 'landscape'. 

![image](https://user-images.githubusercontent.com/5744708/150293551-fa4cadaf-7626-4d54-a6f2-2571fab80f8f.png)

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.
- More issues in this project: https://github.com/orgs/Greenstand/projects/26

jwgatt:
On it.

jwgatt:
Do we want something like this?
![Screenshot from 2022-02-06 16-38-50](https://user-images.githubusercontent.com/79326421/152688671-ba7c3f2e-2043-449d-aa65-31b4d5a12474.png)

jwgatt:
Something like this?

![Screenshot from 2022-02-06 22-05-42](https://user-images.githubusercontent.com/79326421/152701540-3d9e1c70-a8d0-4e6a-a976-b48daca7e0a4.png)

jwgatt:
@dadiorchen Something like this?
![Screenshot from 2022-02-06 22-05-42](https://user-images.githubusercontent.com/79326421/152701687-62695fe6-1158-4a33-8496-65636ed568bd.png)

dadiorchen:
Yes, this is good, I think a smaller corner radius is better.

PR #463, opened 2022-02-07
style: refine planter image

jwgatt:
Fixes issue [#425](https://github.com/Greenstand/treetracker-web-map-client/issues/425)

Some changes made to planter profile image css properties.

Screenshot from 2022-02-06 22-05-42

### DIFF
--- a/src/pages/planters/[planterid].js
+++ b/src/pages/planters/[planterid].js
@@ -100,8 +100,8 @@ export default function Planter({ planter }) {
         variant="rounded"
         sx={{
           width: '100%',
-          height: '688px',
-          borderRadius: 6,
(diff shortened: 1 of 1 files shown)


VWRoli (changes requested):
+1 for smaller border radius, I think we can keep it the same as it was. @jwgatt 

~~Edit: I think the height could actually go down to 400px.~~

EDIT AGAIN: Actually it is 340px in the Figma design so that will be alright. 

Screenshot from 2022-02-08 08-32-26

dadiorchen:
@jwgatt can you resolve the conflict? Then I can merge your code, thanks!

PR #488, opened 2022-02-26
feat: fix the title area drawer mobile

### LINKED ISSUE #406: The mobile drawer should have a fixed title area
As shown in the picture, the red box area is fixed, which means, when you want to scroll the panel (the whole area in the drawer) you need to touch the area under the red box, and, the area in red box can not be scrolled, it will be always fixed on the top.

Also, if you touch and pan the red box, the drawer will be moved.

Screen Shot 2022-01-14 at 5 56 36 AM

---

Some hints:

pierrelstan:
# Description

I created a  custom hook combined with the context API to make the data available globally so that I can access it to the Drawer component, and display the data.

The pages are as children of the Drawer component, and they can not share their data as props or by lifting the state up to the Drawer Component.  Due to that challenge, I used a custom hook combined with Context API to make the data available globally, to have access to it.

Fixed  #406

## Type of change

## Screenshots
Screen Shot 2022-02-26 at 2 26 01 PM

Screen Shot 2022-02-26 at 2 27 26 PM

# Checklist:


src/components/common/Hooks/DrawerHooks.js
```diff
@@ -0,0 +1,34 @@
...
+        treeId,
+        setTreeId,
+        verifiedToken,
+        setVerifiedToken,
+        verifiedTree,
+        setVerifiedTree,
+      }}
+    >
+      {children}
+    </ContextApi.Provider>
```
dadiorchen, line 32: It feels a bit like invasive code for me, Stanly, why you need to use this context.

pierrelstan: to make the data available  for the Drawer title component .

VWRoli (changes requested):
Hello @pierrelstan!
Great idea with the context, did not think about that.

DrawerHooks is not a hook, but a context, it should be DrawerContext, placed into the context folder. And instead of ContextApi, I would name it DrawerContext to be more descriptive. Instead of 'titles, setTitles' I would go with titlesData as it contains data for the title page. 
___
I don't think DrawerTitle should accept any props. I think it should get the props from the context, and handle the logic there. The whole point of the context is not to pass down props.
___
https://github.com/Greenstand/treetracker-web-map-client/blob/5e9c5c20367c5cfbf3a0282251cf18c8da7f0145/src/components/Drawer.js#L173-L200
Here you can just put the DrawerTitle in the StyledBox. No need to wrap it.
```
 <StyledBox 
     sx={{ 
       display: 'flex', 
       justifyContent: 'center', 
       marginBottom: [0, 4], 
       padding: 5, 
     }} 
   > 
     <Puller /> 
<DrawerTitles />
   </StyledBox> 
```
___
The Wrapper components you create are not necessary. I'm guessing you used to use Styled components. With MUI there is no need to do this. Just use a Box or whatever is appropriate, which basically a styled component already.
Same goes for 'IsTitleTreeHideOnMobile'.
___
What I do is when I create a context I also create a hook at the and to access the data from the context. 
```
export const useDrawerContext = () => {
  return useContext(DrawerContext);
};
```
This way I can access the data from everywhere with:
```
 const { titles } = useDrawerContext();
```
This way there is no need to import useContext everywhere, you know.
___
Make sure the title still looks the same as the Figma design margin, colors etc...
___
Add it to the organization page as well. 
___
Just some semantic issues with naming components, nothing to do with the code.
'DrawerHook' as it is a single hook, same with 'DrawerTitle' and not 'DrawerTitles'.
___
Let me know if something is not clear or have some questions!

pierrelstan:
Hello, @VWRoli  thank you for the great review. I will make the changes as soon as possible.

VWRoli (changes requested):
Check the colors and margins again, because they are not correct. 
For example, the Title component color should remain default (#474B4F). The VerifiedBadge margins are too big. 
___
I can just repeat myself. There is no need for creating components that already exist. Subtitle, Title, Wrapper, etc...
Leave them as a simple Typography component. The whole purpose of MUI is that we don't have to create simple components like this. 
It would be appropriate if there would be a more complex component, with multiple divs and other elements.
___
Org page and Planter page title are the same, create one component that can handle both.
___
I don't agree with the Is...HideOnMobile logic. First of all, DrawerTitle is only rendered on mobile. We don't render it on a desktop screen.
How I would approach this is the following:
So we either render a `TreeTitle` or the `Org/Planter Title`. We can use the TreeID for this If TreeID is true render `TreeTitle` else render `Org/Planter Title`. (Which accepts two properties name and CreatedTime).
___
CustomImageWrapper is not present on the planter and Org page. Only on the tree page.
And I don't think you should change anything on the custom Image wrapper component.
I saw that the design changed, but you should focus on the issue at hand. That is the fixed title for the mobile drawer. The new design is a completely different issue. 
___
Don't try to fix everything at once. Only focus on your issue.

pierrelstan:
@VWRoli Thank you for the review. I  made the changes.

dadiorchen:
@pierrelstan thanks for this great job!

Issue #552, opened 2022-05-06
Tiny CSS problem


dadiorchen:
The zoom button is a bit higher than the logo
<img width="542" alt="image" src="https://user-images.githubusercontent.com/5744708/167064370-1a6ada69-bdde-47b8-a36a-ef5694d66f9b.png">

On this page: http://localhost:3000/trees/186735?embed=true
---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.
- More issues in this project: https://github.com/orgs/Greenstand/projects/26

rsmith-github:
I've fixed it on my local version. May I open a pull request?

tranquanghuy0801:
@dadiorchen Seem like this issue has been solved in the linked MR?

dadiorchen:
Yes, thank you

PR #554, opened 2022-05-06
fix: tiny css problem #552

rsmith-github:
# Description

Fixes Tiny CSS Issue #552

Before:
Screen Shot 2565-05-06 at 19 46 10

After:
Screen Shot 2565-05-06 at 19 46 54

### DIFF
--- a/src/components/LayoutEmbed.js
+++ b/src/components/LayoutEmbed.js
@@ -269,7 +269,12 @@ export default function Layout({ children, isFloatingDisabled }) {
             {/* eslint-disable-next-line @next/next/no-img-element */}
             <img alt="zoom-in" src={zoomIn} />
           </Box>
-          <Box onClick={handleZoomOut}>
+          <Box
+            onClick={handleZoomOut}
+            sx={{
+              '& img': { display: 'block' },
+            }}
+          >
             {/* eslint-disable-next-line @next/next/no-img-element */}
             <img alt="zoom-out" src={zoomOut} />
           </Box>
--- a/src/style.css
+++ b/src/style.css
@@ -99,6 +99,7 @@ a:hover {
     opacity: 0;
     transform: translateY(100%);
   }
+
   100% {
     opacity: 1;
     transform: translateY(0);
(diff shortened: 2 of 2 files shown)


rsmith-github:
I found an issue that the css selector wasn't specific enough, I will add it and push the changes again! sorry for any inconvenience.

dadiorchen:
@rsmith-github thank you! Good try, but the way you solve the problem is not so good, as you see, the class name is kand of randomly value, that's auto-generated by material-ui, you need to find the place in the .js file and add the CSS property to that element.

dadiorchen:
And another question, I'm not sure, is the `float` property someone that has been deprecated?

rsmith-github:
Ah ok I understand what you mean! I actually wanted to ask about that as well. In that case I will have another look at the .js file! Would you mind if I try again and commit once I'm done? I'll remove the current css changes. 

In terms of the css float, I don't think it's deprecated but it's more common to use flex box to "justify content"! Thanks for the feedback! 

> @rsmith-github thank you! Good try, but the way you solve the problem is not so good, as you see, the class name is kand of randomly value, that's auto-generated by material-ui, you need to find the place in the .js file and add the CSS property to that element.

dadiorchen:
@rsmith-github great, we heavily use MUI in this app, please read some documents from mui.com, you will get the background of this tech.

>In terms of the css float, I don't think it's deprecated but it's more common to use flex box to justify-content! Thanks for the feedback!

To be honest, I tried to use flex box to solve the problem, but I failed, I don't know why there is a gap under the `zoom-in` button.

rsmith-github:
Hi again, @dadiorchen ! 

I have looked into Material UI and have fixed the css problem in a more dynamic way as you suggested.

I found that by simply changing the 'zoom-out' button's display property to 'block', it fixed the problem. E.g. 

Screenshot 2022-07-25 at 05 13 29

Hopefully this should work. Cheers!

dadiorchen:
@rsmith-github thanks for your contribution!

PR #572, opened 2022-05-19
fix: dev env path resolves #518

### LINKED ISSUE #518: Set up next.js to use relative path for resources
Generally, it's fine when we run the app locally, but when we deploy to the dev env, the next.js is trying to load resources from the root path, but because we run the app under `http://domain/web-map-site/` so this lead to 404 failed to load resources:

image

As shown above, we need the website to request: http://dev-k8s.treetracker.org/web-map-site/xxxx rather than http://dev-k8s.treetracker.org/xxxx

---

Some hints:

okdv:
# Description

simple fix, pushing to dev uses path in .env.dev, which was "". Changed this to /web-map-site so it can load resources correctly.

As a note, this will mean the new local url is [http://localhost:3000/web-map-site](http://localhost:3000/web-map-site). A simple workaround if preferred to keep as it was is to have a .env.local setting the path to "", and thats already ignored by git so it seems like a solid workflow for anyone who prefers to keep the classic url (such as having bookmarks already, etc). Happy to adjust if thats an incorrect assumption though, for example doing this in next instead of in the env. 

Fixes #518 

## Type of change

- [ X ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,1 +1,1 @@
-NEXT_PUBLIC_BASE=""
+NEXT_PUBLIC_BASE=/web-map-site
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "client",
-  "version": "2.0.0-beta.4",
+  "version": "2.0.0-beta.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "client",
-      "version": "2.0.0-beta.4",
+      "version": "2.0.0-beta.5",
       "dependencies": {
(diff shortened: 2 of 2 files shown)


okdv:
re-reading the issue, it seems i missed the part that resolving this in nextjs is what is whats preferred, my bad. Closing this PR and submitting a new one that fixes it in next as opposed to in the env

dadiorchen:
Yes, now we tried to use basePath to solve this problem, but it isn't perfect, if I remember this correctly, we have to `hard code` in some way, like for every link, we need to do like: a=`/{base}/path`

PR #615, opened 2022-05-28
feat: isr - resolves: #571

### LINKED ISSUE #571: Implement page cache via next.js
We need to cache our page, say: http://map.treetracker.org/trees/123 next.js should be having some support for this goal.

---

Some hints:

okdv:
# Description
Implemented Incremental Static Regeneration, see next docs: https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration

Currently only set this up on dynamic routes (trees/orgs/planters), but can implement this on any desired static routes as well. 

**Opening in draft currently as I would like some input from maintainers on the recommended approaches for the below:**
- What static routes would be worth pre-rendering? 
- Should we be caching the map rendering data on the server side as well? I had a hard time finding where exactly that map data gets pulled in at, some background info there would be helpful if this is a desired approach
- Is there an optimal cache revalidation time? I currently have it at 60 seconds, but not for any reason in particular. 

To provide a quick overview on whats happening here:
1. Build time remains the same, no dynamic routes were selected for pre-rendering at build time (although this is possible)
2. When client requests a new dynamic route, it will generate a static page and store on the server side
3. When new, or same, clients request route again-- it will pull in static page, instead of regenerating (i.e. calling DB again)
4. After a period of time (60s currently) the static page is treated as stale and will be regenerated on next call like in step 2 (this timeframe can probably be increased as content doesnt seem to change often)

Fixes #571

## Type of change

New feature (non-breaking change which adds functionality)

## Screenshots

2022-05-27 20 38 58 localhost 023c008dd479
Note the page above contains cache-control header: s-maxage=60, stale-while-revalidate, so caching is successful
image
Note the filesystem in screenshot is from build folder (.next), shows the pages that were statically rendered after being requested

# Checklist:

- My changes generate no new warnings
- New and existing unit tests pass locally with my changes


dadiorchen:
@okdv one suggestion, we can choose to enable/disable this feature by a environment variable, so this is good for debug or better flexicibility.

dadiorchen:
@okdv I just read the doc by next.js, it's a pretty attractive idea, am I correct: the page cache would expire at a specific time, but later request will not get blocked because of expired cache, next.js will trigger a refresh for new coming requests?

okdv:
@dadiorchen 
Can look into the best way to go about controlling this with env. It's worth noting that when running in dev, the page is rebuilt and every request (effectively bypassing cache), if that is suffice for testing/debugging?

And yup basically! After the specified time the cache expires, and the page will be rerendered the next time someone requests it (no auto trigger re render, someone must request page). That will keep the data from getting too stale, and needing a full rebuild to update- but also prevents unnecessary rerenders. If the re render fails for any reason, it will continue showing the previous render. 

The time frame can probably be quite higher i think, as the data doesn't seem to update very often. 

I'll also work to implement this on some static routes also. Off the top of your head, do you know where the map points get called from? That could probably save some time

dadiorchen:
@okdv 

>Can look into the best way to go about controlling this with env. It's worth noting that when running in dev, the page is rebuilt and every request (effectively bypassing cache), if that is suffice for testing/debugging?

Yes, for dev, we can disable the cache for dev, and the env var would be also useful for debugging on prod if we want to disable it temporarily if we ran into some unexpected situation.

dadiorchen:
>The time frame can probably be quite higher i think, as the data doesn't seem to update very often.

It depense, the /trees/[treeid] doesn't change so much, but for the planter's page, and org page, the statistic of trees planted, and species count keeps changing. 

And although the data do not change frequently, once something get modified I guess the end user will always look forward to the update on the page as soon as possible, say, they changed the intro or photo of themselves.

dadiorchen:
>I'll also work to implement this on some static routes also. Off the top of your head, do you know where the map points get called from? That could probably save some time

Sorry, can you explain it a little more? I don't get the question?

okdv:
@dadiorchen 

> Yes, for dev, we can disable the cache for dev, and the env var would be also useful for debugging on prod if we want to disable it temporarily if we ran into some unexpected situation.

This part is proving to be a bit tricky... Ive setup a env variable (`NEXT_CACHE_REVALIDATION_OVERRIDE`) that overrides that time for a simple workaround (i.e. set it to 1 second), but other solutions I can think of would be pretty involved. example, env controlled auto-triggering [this on-demand revalidation](https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration#on-demand-revalidation-beta). Or having 2 versions of each `getStaticProps` page, an ISR one and a CSR one, then redirecting to the appropriate one based on the env. Feel free to take a look on your end as well and see if maybe im missing something more simple.

I also went ahead and played with the values, and implemented this on `top.js` also. But any feedback is welcome, still leaving this in draft for the moment.

okdv:
the fix i went with for overriding this in production is an endpoint that will force revalidation for a desired endpoint, so long there is a token set in the env. 
Would do this by calling /api/revalidate?secret=secret&path=/trees/186737 (where path is the page to revalidate).
If we want this automated, could call the endpoint from pages as they are accessed, while that env var is setup, or just have a function thats called instead of an endpoint. 

Leaving in draft since im not sure if thats a suffice approach, let me know if there is anything different thats preferred

dadiorchen:
Oh, thank you for your great effort on this, @okdv I think the switcher for the cache on prod isn't the most priority, so let's don't get stuck on this, we can figure out this in the future, or maybe we have some way to work around, I guess we can manually reset/clean the cache by restarting or other ways, right? So it's fine if we can not figure out this problem perfectly now.

okdv:
@dadiorchen sounds good, went ahead and scrapped the re-validation bit altogether and move out of draft

dadiorchen:
@okdv what's the situation when we run dev locally? will the page get cached? Or just cache for production/built app

okdv:
@dadiorchen no local dev wont have a persistent cache- it will rebuild the page on each request

dadiorchen:
@okdv perfect!

Issue #521, opened 2022-04-21
Fix a broken Cypress unit test


dadiorchen:
Just try to run the failed test mentioned below and fix it!
https://github.com/Greenstand/treetracker-web-map-client/runs/6106044304?check_suite_focus=true

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.
- More issues in this project: https://github.com/orgs/Greenstand/projects/26

PR #617, opened 2022-06-02
fix: cypress integration tests - resolves #526

### LINKED ISSUE #526: Fix broken Cypress integration test.
Because of some env var problem, our cypree integration tests are broken, please help us fix them.
Here is an example showing the error message.
https://github.com/Greenstand/treetracker-web-map-client/runs/6139006376?check_suite_focus=true

---

Some hints:

okdv:
# Description

There were a number of errors at play here, so a few fixes all over, one left to fix still. 

The env variables in cypress were broken, as these were previously getting pulled from .env, but not any longer. the dotenv package used doesnt take other .env's into consideration- resulting in many errors. I have added the env variables via cypress.json instead, however this means cypress tests will be entirely failing if the url path in cypress.json isnt matching the env url path (e.g. /web-map-beta/demo), so thats a quirk. 

Also a random linting error pertaining to Drawer.js that i fixed by assigning a variable equal to the ref. functionally the same but satisfies lint. 

Lastly there is still an error surrounding [organizationid].js, it seems the data isnt getting mock pulled in so the name/etc never populate. Resulting in the remaining integration test error. I havent given it much of a look yet, but so far no luck in a few things a tried, so any ideas there is helpful
image

Leaving in draft now while i fix that last error, in meantime feel free to let me know of any preferences or requests surrounding this.

Fixes #526

## Type of change
- Bug fix (non-breaking change which fixes an issue)
- This change requires a documentation update -- the cypress.json env quirk is worth documenting probably. Its easy to assume it would pull env data, but unfortunately doesnt pull env data not in .env explicitly. 

# How Has This Been Tested?

- Cypress integration

# Checklist:

- I have commented my code, particularly in hard-to-understand areas
- [ No but likely will need to mention it if this env change is kept] I have made corresponding changes to the documentation
- My changes generate no new warnings


dadiorchen:
For the dotenv problem, do you mean we can not load env like .env.production? Could we do it manually, like: `dotenv.config({path: '.env.development'});`?

https://github.com/Greenstand/treetracker-web-map-client/blob/c32baf03830081ea91ad49eda803d4afa302ec7e/cypress/plugins/index.js#L7

For the mock problem, we used some extra lib to make the development smooth, like auto-refresh the test when save file. So it's not so straightforward, I also need some time to dig into it, it was accomplished by @QuaidBartolomei , @QuaidBartolomei I remember we can do a quick mock on the test file, right? Can you help us to find an example of that?

okdv:
went ahead and just updated the test to match the data thats actually expected to display since its not coming from that api, and fixing that seems a little out of the scope of this issue :smile: 

aside from that I verified all the integration tests are passing now in a dev env, so this should be all good to go. submitting this PR as ready but let me know if theres anything else I should look at

dadiorchen:
>As for the mock data, I looked at it a bit more but still no luck in finding where its coming from. I see that the bad data is being returned here https://dev-k8s.treetracker.org/query/organizations/1 but as far as where that api is from, im not sure...

If I remember correctly, the goal of the nock is to mock the data from the server API, so the test would not rely on the external service, in that way, the test is fragile because the API is unstable, and the test should fail because of the change of external env.

dadiorchen:
Please hold on for a while on this, I can join in and run your branch on my computer to check the situation.

okdv:
@dadiorchen sounds good, leaving it as is. also just noticed it still shows it failed the integration test from gh actions still, but only for a 500 error on top page now. seems to be passing okay locally so thats interesting, but i'll let you take a look when you get the chance 
image

dadiorchen:
@okdv could you go forward to solve this: https://github.com/Greenstand/treetracker-web-map-client/issues/521

okdv:
@dadiorchen sure thing!

dadiorchen:
Some findings: https://greenstand.slack.com/archives/CA0ALV5S8/p1655188570363239

okdv:
@dadiorchen got the 500 error on the top page fixed, as well as the missing info on the individual pages like [organizationid], thanks for the additional context on that 

a better long-term solution could be put in place for the nockRoutes fix, but for the time being it seems like adding specific routes should do just fine.

okdv:
hmm i actually see its still failing on the top page... not quite sure why, doing these tests locally using nock they were all passing. may have to dive into the cypress side a bit more, not sure its coming down to nock at this state

okdv:
made the nock routes persist after their first call, was really the only thing i could think of as a fix for that last error. still running locally with no errors on my side. 

if it fails again, may need another set of eyes to find out how to reproduce this locally. can probably track this down from there.

dadiorchen:
@okdv cool, let me checkout your branch and test.

dadiorchen:
@okdv I will close this PR because I have copied your code and put into another one to solve this problem totally, as I mentioned on Slack

dadiorchen:
Thanks, you did help solve this.

Issue #431, opened 2022-01-20
A mechanism to modify the them in a preview sandbox


dadiorchen:
We are going to provide the greenstand map to other partners, so they can embed the map to their products, like websites, and mobile apps.

So we plan to offer an online sandbox to let our partners customize the appearance of the map app. Kind of like we do when we want to custom the theme of our Twitter profile.

At this early stage, don't need to consider the data storage and permission, just implementing a client sandbox, is good enough.

![image](https://user-images.githubusercontent.com/5744708/150309781-86579be2-116f-4ac0-b144-c01039507fbd.png)

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.
- More issues in this project: https://github.com/orgs/Greenstand/projects/26

dadiorchen:
This is a use case for this kind of feature:
https://www.treetown.finorx.com/map
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/5744708/156274461-695bfcea-3574-4100-88c9-26588bb4da86.png">

dadiorchen:
Eventually, we want to implement something similar to https://bareynol.github.io/mui-theme-creator/

dadiorchen:
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/5744708/175754668-6a9db08e-11a5-4f8e-a018-421be5802a60.png">

RubenSmn:
@dadiorchen I would like to work on this one, can you fill me in on the details? Is there a design in the figma?

dadiorchen:
I have done some work on this PR, https://github.com/Greenstand/treetracker-web-map-client/pull/644 is about login to dashboard and edit the theme, and the way to switch theme via localstorage, and we also created a new microservice to handle the map config for this. 
So now we are going to do an experiment to implement a visualized way to set the theme, as you see in the picture above. 
The UI for this is not so important, we just open this feature for the operator of greenstand. so I want to go forward to implement the function first, then back to improve the UI.
Maybe we can arrange a quick call on slack, and explain and show to you furtherly.

dadiorchen:
[...]

PR #654, opened 2022-06-30
feat/theme playground

RubenSmn:
# Description
Add theme customization playground

Implements #431

## Type of change

# How Has This Been Tested?

- [ ] Cypress integration
- [ ] Cypress component tests
- [ ] Jest unit tests

# Checklist:


RubenSmn:
### Current state
- can customize colors, updates the values across the editor
- color thumbnail to quickly see the colors
- can reset *whole* theme to default
- can enable auto reload, reloads when value changes

### Screenshot 
#### Full
image

#### Editor
image

dadiorchen:
@RubenSmn this is awesome, are you implementing the component by yourself or you are using the existing solution?

RubenSmn:
I'm building the components myself and using the https://bareynol.github.io/mui-theme-creator/ as a design reference/example.

RubenSmn:
### Current state
- can customize typography
- can directly see changes in thumbnail
- prop rules to validate user input

### Screenshot
image

RubenSmn:
Editing props works both ways! You can edit the from `InputFields` or from the textarea with all the styles.

RubenSmn:
### Updated state:
- theme editor will use current custom theme if present
- theme editor will start in current theme mode
- fixed spacing not being passed to custom theme

*no  UI updates*

dadiorchen:
@RubenSmn Just share some my experience, the font family relys on two factor: the font installed locally on the user's computer, manually dynamically load the font from web site via some meta indicator, like <link> so if we don't have font supported by these two approaches, then the font dont's work for current user/browser.

Basically we can manually load various fonts from fronts.google.com. 
Also, the font-weight matters, usually we just load the weight of that font we need, to save network traffic. 

And, there are some weird things I still don't figure out, is that even we download the same font and weight from the internet, I found sometimes they have minor difference with the Figma design, I don't know the reason, just a head up for you in case you run into some problem.

RubenSmn:
@dadiorchen just implemented the basic font customization.
Now looking to store the used fonts in the local storage `themeObject`. The stored fonts should also be loaded when the theme is loaded from local storage. When that is working I will be looking at optimizing getting the fonts with just the required font-weights.

dadiorchen:
@RubenSmn here are some relevant code:

https://github.com/Greenstand/treetracker-web-map-client/blob/75528cba577fca517aed4e25ec4643d13db7d443/src/pages/_document.js#L6-L24

RubenSmn:
@dadiorchen I'm currently loading the fonts with the WebFontLoader [here](https://github.com/Greenstand/treetracker-web-map-client/pull/654/commits/2b2dfb6ecc27d38f2ef66e041317ae15fc4b43ec#diff-5bfabd7c7bac1d5b79ed9847492aecdaa2a72985eb1fb47536e9b76e67fcd9b6R130-R154)

dadiorchen:
@RubenSmn that looks great

RubenSmn:
### Updated state:
- localStorage stored theme has only one theme, using dark/light versions only on props that require them. (palette/components)
- custom fonts are optimized for the web app by adding only the used fonts

*no UI updates*

RubenSmn:
### Updated state
-  every playground component has a test

*no UI updates*

dadiorchen:
> ### Updated state
> * every playground component has a test
> 
> _no UI updates_

@RubenSmn I recommend the Test Driven Development, in this way, you build tests first, then the code, I rely on this heavily, and Cypress also brings great convenience to practice TDD, comparing to the FB recommended lib testing-library, because you can code watching the UI.
BTW, I think the principle for web map client in terms of the test coverage is no need for 100% unit test covering every component, if they already are covered by the integration test.

RubenSmn:
@dadiorchen thanks for sharing your knowledge! I will try to apply this in the future, now that I know a little more about cypress!

Issue #707, opened 2022-07-30
Use the route to control the embed behavior


dadiorchen:
Currently, we use this way (query parameter) to control the embed behavior:

http://localhost:3000/planters/940/trees/186734?embed=true

This is annoying because we need add this parameter for every URL when we navigating on the app.

We need to consider use route to control this behavior:

http://localhost:3000/m/planters/940/trees/186734

Probably we need to use next.js rewrite to implement, but any possible solution is welcome!

dadiorchen:
We were thinking about use cooike to implement it, but in this way, we can not let user open the embed map at the first time, right?

http://localhost:3000/planters/940/trees/186734?embed=true

okdv:
I would agree with the cookie solution in #591, as opposed to a next.js rewrite. It's a little more involved, but i think this could work well by creating some routes dedicated to setting and getting the cookie. 

- create /getCookie and /setCookie (as examples)
- in _app check /getCookie
- - if /getCookie finds no cookie, uses /setCookie to set default and returns that
- - if /getCookie finds existing cookie, returns that
- assign /getCookie value to same nextExtraIsEmbed prop and pass it along to page thats being requested so not much work needed in the individual pages/components 

then when a user hits the map embed/fullscreen toggle button, it can call /setCookie and toggle it, and refresh the page, running _app and retrieving new cookie value from /getCookie, etc- cycle continues

if that sounds good, im happy to take a crack at it :smile:

dadiorchen:
@okdv I'm also dislike this solution, it seems not so RESTFul style, the `emebed` is a state of UI rather than a resource, so I don't want to put it into the path rather than query parameter.

But when I re-think the requirements, you know, for the partner of greenstand, say, https://sadu.app/#everyone they want to integrate our map into their app (as a browser view) in this case, we need totally remove the 'un-embed` button, how do we handle this case? So I was thinking use http://domain/m/... to bring a total embed web map app.

Then back to our `embed/un-embed` button, the cookie/localstorage approach is a good solution to avoid cumbersome query parameters.

Maybe we need to implement them all...

what do you think?

dadiorchen:
another question: cookie vs localestorage:

I prefer local storage, it's easy to manipulate, no server-data-sending, so don't need to ask permission of user, last one: we already used it in the project.

@okdv @RubenSmn please comment

RubenSmn:
[...]

PR #777, opened 2022-08-10
Context/Global hooks

okdv:
# Description

As discussed in slack, a central place for reading details from context or other global usage settings (is mobile, is embed, etc)

Opening in draft to make sure all are on same page with how it becomes structured. Only real change to plan I made was creating a contextHooks (theme, playground) and globalHooks (embed, mobile, etc) to separate the two major types. This was done to prevent context from requiring hooks, while hooks also require context. 

There is probably a bit more cleaning up that could be done in general if we move forward with this sort of consolidation. 

passes cypress unit testing
confirmed localStorage and context are still held across web app as expected


RubenSmn:
Maybe we could make the `contextHooks` and `globalHooks` into a folder and create seperate files for the hooks and export them from an `index.js`. I think long term if we get more hooks this would be cleaner. What do you think? @okdv

src/hooks/globalHooks.js
```diff
@@ -0,0 +1,48 @@
...
+  }, [key, value]);
+
+  return [value, setValue];
+};
+
+// Determine if mobile based on theme breakpoints
+export const useMobile = () => {
+  const theme = useTheme();
+  return useMediaQuery(theme.breakpoints.down('sm'));
+};
```
dadiorchen, line 42: If we want to control the mobile status, say, in the customizable web map, we put the whole app into a box and let them render in the mobile size, so in this case, maybe we need somewhere to put the states, and the useMobile will return different value based on the status?

dadiorchen: @RubenSmn I mean if some day we need to consider do not using iframe, and use component to embed the sandbox instead.

RubenSmn: Hmm yes true. I think if we'll use an component to render the preview we still need an container and change the height an width of it in order to create a mobile view.
I am not sure is there is a way we could force is by an state.

okdv: so we would like a way to override the behavior of useMobile? maybe we use the URL, similar to useEmbed, and save that info to context/state only instead of localstorage?

RubenSmn: Yes something like that

okdv: im about to work another issue #707 relating to url parameters in general, so i'll definitely include that

dadiorchen: @okdv I'm not requiring features, I am discussing the solution, so no need to do it if not current necessary, just keep this in mind

dadiorchen: Now we are using iframe and it works fine

RubenSmn (approved):

dadiorchen:
@okdv huge progress!

PR #778, opened 2022-08-10
chore: ignore swap files

RubenSmn:
# Description
Ignores swap files for git
For the vim users : )


dadiorchen:
:D if you use neovim, it will not put the swap file into your work folder.

RubenSmn:
Oh, that is better!

PR #796, opened 2022-08-12
feat: URL embed control

### LINKED ISSUE #707: Use the route to control the embed behavior

okdv:
# Description

Changed the approach to this slightly, feel free to let me know if it misses any use-cases. But I effectively made embed=true as the way to force embedding at all times instead of making a new param (and hide un/embed button)- for things like partners iframes/webviews/etc. This embed=true will follow any link clicks, etc.

Actual users of the website will not use embed=true, they will just control embed using the un/embed button, and its all done through localStorage/state- URL never reflects embed=true for them, allowing us to dedicate it to those partners as mentioned.

One thing to note, there are a number of components where URL parsing/manipulation takes place, I didnt update each one yet as i wanted to make sure my fix fits our case before doing extra work. If this fix is okay, I will add another commit for cleaning this everywhere. 

Fixes #707 

## Type of change

- [x] New Feature

# How Has This Been Tested?

- [x] Cypress integration
- [x] Cypress component tests
- [x] Jest unit tests

# Checklist:

### DIFF
--- a/src/components/LayoutEmbed.js
+++ b/src/components/LayoutEmbed.js
@@ -5,6 +5,7 @@ import Drawer from '@mui/material/Drawer';
 import dynamic from 'next/dynamic';
 import React from 'react';
 import SearchFilter from './SearchFilter';
+import useEmbed from '../hooks/useEmbed';
 import LogoIcon from '../images/greenstand_logo.svg';
 import MinIcon from '../images/min.svg';
 import ZoomIn from '../images/zoom-in.svg';
@@ -56,6 +57,7 @@ export default function Layout({
 }) {
   const [isDrawerOpen, setIsDrawerOpen] = React.useState(true);
   const [toggleButtonPosition, setToggleButtonPosition] = React.useState(0);
+  const isEmbed = useEmbed();
   // const { _classes } = useStyles();
   const mapContext = useMapContext();
   function handleFullScreen() {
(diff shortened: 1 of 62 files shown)


dadiorchen:
I agree with you Otho, good observation!

okdv:
@dadiorchen awesome, will get the cleanup done and submit for review

src/components/LayoutEmbed.js
```diff
@@ -177,24 +179,25 @@ export default function Layout({
...
-        }}
-      >
-        <SvgIcon
-          component={MinIcon}
-          inheritViewBox
-          sx={{ height: 52, width: 52 }}
-        />
-      </Box>
-
+      {isEmbed || (
```
RubenSmn, line 182: Should this be an OR? Now it will always render the elements

src/pages/index.js
```diff
@@ -19,7 +20,7 @@ export default function Homepage() {
         <meta name="viewport" content="width=device-width,initial-scale=1.0" />
         <title>Treetracker by Greenstand</title>
       </Head>
-      <Home />
+      {nextExtraIsEmbed || <Home />}
```
RubenSmn, line 23: Same OR here

src/components/LayoutEmbed.js
okdv: since logical operators short circuit, it should only return the component if isEmbed is falsey

src/components/LayoutEmbed.js
```diff
@@ -177,24 +178,25 @@ export default function Layout({
...
-        }}
-      >
-        <SvgIcon
-          component={MinIcon}
-          inheritViewBox
-          sx={{ height: 52, width: 52 }}
-        />
-      </Box>
-
+      {useEmbed || (
```
dadiorchen, line 181: @okdv is `useEmbed` a function, if true, why you just use function name rather than executing it for checking?

src/components/SearchFilterMobile.js
```diff
@@ -5,6 +5,7 @@ import { useRouter } from 'next/router';
 import React from 'react';
 import Link from './Link';
 import Filter from './common/Filter';
+import { useEmbed } from '../hooks/globalHooks';
```
dadiorchen, line 8: @okdv don't need to worry about the stuff related to search/filter, we are not using it, and we will re-implmenent the logic in the future.

dadiorchen:
@okdv I will merge your PR first, because I run into some problem with the path refresh, maybe your PR can solve it

src/components/LayoutEmbed.js
RubenSmn: Why are we checking for an embed on the embed layout?

okdv: @dadiorchen  ah yup you are correct, it should be treated as a function. just updated that and submitting a PR for it 

@RubenSmn The naming is a little ambiguous so perhaps that should be updated to make things less confusing. But basically useEmbed is checking for the embed=true parameter and doesnt look at the localStorage. This way we can hide the un/embed button when embed=true is a parameter, but still show when embed is only true in localStorage (i.e. a user has clicked the embed/expand map button, so we want to leave the untoggle button)

PR #806, opened 2022-08-15
fix: species component no key

RubenSmn:
# Description
Fixes warning of component having no key

## Type of change

# Checklist:

### DIFF
--- a/src/components/TreeSpeciesCard.js
+++ b/src/components/TreeSpeciesCard.js
@@ -21,7 +21,7 @@ function TreeSpeciesCard(props) {
       sx={{
         borderRadius: 4,
         overflow: 'hidden',
-        mb: [2, 4],
+        my: [2, 4],
       }}
     >
       <Grid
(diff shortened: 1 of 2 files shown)


src/components/TreeSpeciesCard.js
```diff
@@ -21,7 +21,7 @@ function TreeSpeciesCard(props) {
       sx={{
         borderRadius: 4,
         overflow: 'hidden',
-        mb: [2, 4],
+        my: [2, 4],
```
dadiorchen, line 24: @RubenSmn let's remove these margin from the component and let the client to deal with the spacing, in this way, we make the component more re-useable.

dadiorchen: So, remove margin from compnent and put it into the invoker of this component

RubenSmn:
@dadiorchen took another approach and fixed it!

src/pages/organizations/[organizationid].js
```diff
@@ -367,14 +367,14 @@ export default function Organization(props) {
                 ))}
                 {/* Placeholder, remove after API fixed */}
                 {organization?.species?.species?.map((s) => (
-                  <>
+                  <React.Fragment key={s.name}>
```
Comment (dadiorchen) [inline src/pages/organizations/[organizationid].js:370]: I guess this is the same as: `<></>` ? right?

Comment (RubenSmn) [reply src/pages/organizations/[organizationid].js:370]: Yes but this way you can pass the key

PR #808, opened 2022-08-15
feat: map can be now loaded by bounds

### LINKED ISSUE #754: Should be able to load the map by `bounds`
If the url includes `bounds` parameter, then the map should be load by the bounds:

http://localhost:3000/planters/940?bounds=38.79890441894532,-5.565415753407831,39.29328918457032,-5.069057826784033

As shown above, when open the page, the map should display the view according to the bounds=lat,lon...

The map core (Map object) can be set by bounds, so we just need to set it when we load the page.

Just like our online version: https://map.treetracker.org/?&bounds=13.719177246093752,14.599531422729925,17.674255371093754,16.70986293320658

Mohmn:
# Description

Fixes #754

## Type of change

## Screenshots

# How Has This Been Tested?

- [ ] Cypress integration
- [ ] Cypress component tests
- [ ] Jest unit tests

# Checklist:

### DIFF
--- a/src/pages/organizations/[organizationid].js
+++ b/src/pages/organizations/[organizationid].js
@@ -110,14 +110,17 @@ export default function Organization(props) {
       // manipulate the map
       const { map } = mapContext;
       if (map && organization) {
-        // map.flyTo(tree.lat, tree.lon, 16);
-        map.setFilters({
-          map_name: organization.name,
-        });
         // TODO why I must try/catch this?
         try {
-          await map.loadInitialView();
-          map.rerender();
+          if (router.query.bounds) {
+            await map.goToBounds(router.query.bounds);
+          } else {
+            map.setFilters({
+              map_name: organization.name,
+            });
+            await map.loadInitialView();
+            map.rerender();
+          }
         } catch (e) {
           log.error('rendering map:', e);
         }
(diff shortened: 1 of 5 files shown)


Mohmn:
@dadiorchen run these changes in ur local first to check if it is working properly or not

src/pages/organizations/[organizationid].js
```diff
@@ -110,14 +110,17 @@ export default function Organization(props) {
...
-          map.rerender();
+          if (router.query.bounds) {
+            await map.goToBounds(router.query.bounds);
+          } else {
+            map.setFilters({
+              map_name: organization.name,
+            });
+            await map.loadInitialView();
+            map.rerender();
+          }
```
Comment (dadiorchen) [inline src/pages/organizations/[organizationid].js:123]: @Mohmn setFilter will reset the state of the map, to load expected map, like: load organization map above, and even we got bounds from query parameter, I think we should stop to load the org map? 

According to your code, we just jump to the map area, and without loading the correct state of the map? (by setFilter)

Comment (dadiorchen) [reply src/pages/organizations/[organizationid].js:123]: You can test it in this way: 
1. Open an planter map: https://dev-k8s.treetracker.org/web-map-beta/demo/planters/940
2. Then move the map a little bit, so the `bounds` shows up in url
3. Copy the URL and open in browser, then the result should be:
   1. The map jump to the bounds
   2. The map show the planters tree/clusters.

PR #826, opened 2022-08-17
fix: useEmbed in LayoutEmbed

okdv:
# Description

Fixing the bad reference of useEmbed from my last PR #796 that @dadiorchen caught 

@RubenSmn also pointed out that the name sharing between embed parameter and localStorage is ambiguous and confusing, should I throw this into this PR? Maybe renaming useEmbed hook to something like useEmbedParam?


RubenSmn:
`useEmbed` is a fine name for the hook.

The thing was that you were checking whether or not the embed was true in the `EmbedLayout` and the layout itself only shows if `embed` is `true` :)

okdv:
@RubenSmn ah ok I can probably submit for review then, figured we could change the name while we're at it to prevent further confusion since that call was deliberate, it just reads to be pointless. It covers the use case of leaving the un/embed button visible when users are in embed mode, but do not have embed=true present in the url params (ie they got their via embed toggle button)

RubenSmn (approved):

PR #833, opened 2022-08-19
docs: add figma design on issue template

RubenSmn:
# Description
Add figma design link on issue template to provide more infomation


dadiorchen:
@RubenSmn nice!

PR #836, opened 2022-08-19
ISR/Cache for /top

### LINKED ISSUE #797: Set next.js to pre-cache pages when starting
We need pre-cache some pages because they are pretty time consuming, we don't want to let our end user waiting for the page, currently, these page need to be pre-cached:

`/top`

Please also consider a way to opt-out this feature, for example, we don't want to do it when we developing locally

okdv:
# Description

Using ISR for /top, pre-renders /top at build time
Also hot fix for undefined planter image call

for ISR:
- does not run at development build time, only production
- in development, /top is rebuilt on each request
- in production, /top is rebuilt on next call after successful re-render, no more than every 60 seconds (time is flexible obviously)

Fixes #797

## Type of change

# How Has This Been Tested?

- [x] Cypress integration
- [x] Cypress component tests

# Checklist:

### DIFF
--- a/src/pages/top.js
+++ b/src/pages/top.js
@@ -222,7 +222,6 @@ export async function getStaticProps() {
       organizations,
     },
     revalidate: 60,
-    fallback: 'blocking',
   };
 }
 
--- a/src/pages/top.js
+++ b/src/pages/top.js
@@ -197,21 +197,21 @@ function Top({ trees, planters, countries, organizations }) {
   );
 }
 
-export async function getServerSideProps() {
+export async function getStaticProps() {
   const [trees, countries, planters, organizations] = await Promise.all([
-    getFeaturedTrees(), //
+    getFeaturedTrees(),
     process.env.NEXT_PUBLIC_COUNTRY_LEADER_BOARD_DISABLED === 'true'
       ? []
       : getCountryLeaderboard(),
     (async () => {
       const data = await utils.requestAPI('/planters/featured');
       log.warn('planters', data);
-      return data.planters;
+      return [data];
     })(),
     (async () => {
       const data = await utils.requestAPI('/organizations/featured');
       log.warn('organizations', data);
-      return data.organizations;
+      return [data];
     })(),
   ]);
   return {
@@ -221,6 +221,8 @@ export async function getServerSideProps() {
       planters,
       organizations,
     },
+    revalidate: 60,
+    fallback: 'blocking',
   };
 }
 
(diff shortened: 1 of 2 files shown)


okdv:
Should we add any other pages into this or leave at just /top for now? if just top, can move into ready for review

src/pages/top.js
```diff
@@ -197,21 +197,21 @@ function Top({ trees, planters, countries, organizations }) {
   );
 }
 
-export async function getServerSideProps() {
+export async function getStaticProps() {
```
dadiorchen, line 200: @okdv can you explain this a little more? What would be the behavior using this way?

src/pages/top.js
```diff
@@ -197,21 +197,21 @@ function Top({ trees, planters, countries, organizations }) {
...
-    getFeaturedTrees(), //
+    getFeaturedTrees(),
     process.env.NEXT_PUBLIC_COUNTRY_LEADER_BOARD_DISABLED === 'true'
       ? []
       : getCountryLeaderboard(),
     (async () => {
       const data = await utils.requestAPI('/planters/featured');
       log.warn('planters', data);
-      return data.planters;
+      return [data];
```
dadiorchen, line 209: Why do we need to change to this, and as a result, do we need to also change the way to visit this in the client?

src/pages/top.js
okdv: @dadiorchen sure thing

getServerSideProps calls that data at load time from the client side (to my understanding, what we are trying to avoid for this page)

getStaticProps generates a static, server side rendered page at build time. This static data will be re-rendered based on the rule provided in the revalidation rule

there is a lot more in depth info in the nextjs docs though, so i'll defer to that https://nextjs.org/docs/basic-features/data-fetching/overview

src/pages/top.js
okdv: this was unrelated to this issue, but prevented me from loading the page/confirming correct implementation. If i remember correctly, /planters/featured was returning a single object of a planter, but we were expecting an object with array of planters in it, causing errors as data.planters does not exist but is necessary for rendering the page. 

I'll doublecheck in a bit to make sure im remembering it correctly, but feel free to let me know if youre seeing something else

dadiorchen: Oh, I think it's fine, we will see and if there are problem emerging, we can open issue then.

src/pages/top.js
dadiorchen: @okdv perfect

PR #837, opened 2022-08-19
refactor: extract information tags

RubenSmn:
# Description
Refactor the information tags to a seperate component.
Create a new data structure to generate the tags.

For now only refactored the `trees` page. Will do the other pages if this approach gets approved.

## Type of change

# Checklist:

### DIFF
--- a/src/components/common/TagList.js
+++ b/src/components/common/TagList.js
@@ -1,0 +1,44 @@
+import { Box, SvgIcon } from '@mui/material';
+import TreeTag from './TreeTag';
+
+function TagList({ tagListInfo }) {
+  const head = [];
+  const tail = [];
+
+  tagListInfo.forEach((tag) => {
+    const tagComponent = (
+      <TreeTag
+        key={`${tag.title}-${!tag.value ? tag.error : tag.value}`}
+        TreeTagValue={!tag.value ? tag.error : tag.value}
+        title={tag.title}
+        icon={<SvgIcon component={tag.icon} />}
+        disabled={!tag.value}
+        subtitle={tag.value ? tag.subtitle : null}
+        link={tag.value ? tag.link : null}
+      />
+    );
+
+    if (!tag.value) {
+      tail.push(tagComponent);
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


src/components/common/TagList.js
```diff
@@ -0,0 +1,44 @@
...
+      <TreeTag
+        key={`${tag.title}-${!tag.value ? tag.error : tag.value}`}
+        TreeTagValue={!tag.value ? tag.error : tag.value}
+        title={tag.title}
+        icon={<SvgIcon component={tag.icon} />}
+        disabled={!tag.value}
+        subtitle={tag.value ? tag.subtitle : null}
+        link={tag.value ? tag.link : null}
+      />
+    );
```
dadiorchen, line 19: @RubenSmn I will say I don't endorse this component, I know there are some redundant in the code for the original version, and we shouldn't repeat ourselves in pages, but I think:

1. This dedicated tag list is not a common case, I can not image too many cases to repeat this pattern, so it is not so useful.
2. Function/class/component pay cost, you need extra code to define it, you need to maintain then, you need to define the interface, also more important, you need to be enough flexible to adopt different cases, any change on the interface has impact/risk for all your clients. So do or do not extract a component, that's a question about trade-off.
3. Good component need a good domain/model so it is highly re-usable and solid, in this case, the domain(requirement) is not so clear, a list with head tags and tail tags? In your code, the way to decide the type is by `!tag.value` but that's easy to break, say, I change the rule: I will put some `value=unverified` to the tail, so we need to change the code, I think the root is that we don't have a solid domain for this component, which make it fragile.
4. Component brings more complexity, even I am familiar with the original code, I need to take some time to understand the usage of the component.

So I don't support extracting this component.

RubenSmn: @dadiorchen I understand, thank you for the feedback! I do think we should find another solution to display the information tags to clean up the `id pages`

dadiorchen: I understand your worries, yes, let's see if there are better way to streamline the code

PR #859, opened 2022-08-24
fix: isr usage of msw

### LINKED ISSUE #857: Failed to use pre-cache feature by next.js
@okdv I failed to release the new version to prod env, please help me analyze, this is the error: https://github.com/Greenstand/treetracker-web-map-client/runs/7990205499?check_suite_focus=true

Seems we can not build next now, can you check it on your side?

okdv:
# Description

Related to my previous PR for using ISR on the /top page

Basically issue was in dev env, we are using MSW. MSW handlers were designed incorrectly resulting in them returning a single planter/org object when requesting /planters/featured or /organizations/featured (was picking up featured as the ID). However, prod api returns object with an array of planters/orgs inside of it.

In my last PR, I was relying on MSW data to make change that returned [data] instead of data.planters or data.organizations (since mock data returned an object and we were expecting an array).

To fix this, I updated MSW to return data matching that of the prod api and updated the /top ISR usage so we can take advantage of caching. This also involved updating the image key name for the planters.

Fixes #857

## Type of change

# How Has This Been Tested?

- [x] Cypress integration
- [x] Cypress component tests

# Checklist:

### DIFF
--- a/doc/examples/planters/940.json
+++ b/doc/examples/planters/940.json
@@ -2,7 +2,7 @@
   "id": 940,
   "first_name": "Sebastian ",
   "last_name": "Gaertner",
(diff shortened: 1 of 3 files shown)


okdv:
to clarify, this PR also readds ISR abilities from my last PR #836 but now actually working with the prod api

PR #862, opened 2022-08-26
mainworkflow

dadiorchen:
- feat: add link for species
- fix: can click tree on planter page
- fix: treepage map can not load correctly
- fix: org map
- fix: disable country boad
- feat: tree page enter wallet
- Revert "fix(drawer): not able to scroll to bottom content"


dadiorchen:
@RubenSmn can you do some research on this? I reverted your commit 9fbeb59160e9ea7cd979e2d2807027a3a700f0d7, seems this is the reason that do some interference in the drawer experience?

RubenSmn:
@dadiorchen I will have a look!

PR #865, opened 2022-08-26
feat: avoid drawer move when touch

dadiorchen:


RubenSmn:
Scroll to the bottom is not working again

image

PR #866, opened 2022-08-26
fix: removed sx prop from CustomeCard.js

### LINKED ISSUE #843: Too big size of the icons on mobile
http://localhost:3000/wallets/eecdf253-05b6-419a-8425-416a3e5fc9a0

Image

Maybe we need to find a good way to reuse/reduce-maintainness of these cases

PhiCygni:
# Description

[comment]: # 'resolves #843 The SvgIcon in CustomCard.js no longer uses the sx prop being passed in and instead is fixed to 32px for breakpoint below sm and 52px for breakpoint above sm. Senior devs suggested the change in the issue description https://github.com/Greenstand/treetracker-web-map-client/issues/843'

Fixes #843

## Type of change

## Screenshots

mount Card -- CustomCard (1)
mount Card -- CustomCard

# How Has This Been Tested?

- [x] Cypress component tests

# Checklist:

### DIFF
--- a/src/components/common/CustomCard.js
+++ b/src/components/common/CustomCard.js
@@ -11,7 +11,7 @@ const useStyles = makeStyles()((theme) => ({}));
 function CustomCard(props) {
   const { classes } = useStyles(props);
   const theme = useTheme();
-  const { iconURI, title, text, handleClick, disabled, tooltip } = props;
+  const { iconURI, title, text, handleClick, disabled, tooltip, sx } = props;
 
   return (
     <Box
@@ -46,6 +46,7 @@ function CustomCard(props) {
         >
           <SvgIcon
             sx={{
+              ...sx,
               height: [32, 52],
               width: [32, 52],
               fill: 'none',
--- a/src/components/common/CustomCard.js
+++ b/src/components/common/CustomCard.js
@@ -46,9 +46,9 @@ function CustomCard(props) {
         >
           <SvgIcon
             sx={{
-              ...sx,
               height: [32, 52],
               width: [32, 52],
+              ...sx,
               fill: 'none',
             }}
             component={iconURI}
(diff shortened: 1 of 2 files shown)


RubenSmn:
@PhiCygni great effort!
But we cannot just remove this `sx` prop due to the fact that it is widely used; for example https://github.com/Greenstand/treetracker-web-map-client/blob/beta/src/components/ImpactSection.js#L48-L61 and probably every other place we use the CustomCard component
We also need to be able to change the `sx` for the color of the icon. I do like the fix of putting a standard height/width in the SvgIcon!

PhiCygni:
@RubenSmn Very understandable. I will look at ImpactSection.js and see how it works with the sx prop. Thanks for the feedback.

RubenSmn:
Oh and for the issue to be referenced you need to type the full `#843` instead of `# (843)`

PhiCygni:
I see, I was wondering if I should remove those brackets. Thanks

PhiCygni:
The SvgIcon now still has the sx prop. In the Cypress test I tested the different sx prop configurations I saw in the real app to make sure it works.

mount Card -- CustomCard (1)
mount Card -- CustomCard

src/components/common/CustomCard.js
```diff
@@ -47,6 +47,8 @@ function CustomCard(props) {
           <SvgIcon
             sx={{
               ...sx,
```
RubenSmn, line 49: Could you move the sx beneath the height and width? That way we can override the default height/width if we need to, some icons need to be smaller/larger depending on the case

RubenSmn (changes requested):

src/components/common/CustomCard.js
PhiCygni: I see sure, making the change now.

PhiCygni:
Moved the sx below the height and width.

mount Card -- CustomCard (1)
mount Card -- CustomCard

RubenSmn:
@PhiCygni this looks great! The first two icons on mobile are still a bit too big. Could you fix this. I think they are also used in the `src/pages/wallets/[walletid].js`

dadiorchen:
I will close this PR because it has been solved

PR #896, opened 2022-09-05
refactor: render order based on disabled prop

RubenSmn:
# Description
Improvement on #837
Add `TagList` which renders the tags based on the disabled value

## Type of change
Refactor

## Screenshots
Top -  render with `TagList`
Bottom - old with `if-else`
image

# Checklist:

### DIFF
--- a/src/components/common/TagList.js
+++ b/src/components/common/TagList.js
@@ -1,0 +1,30 @@
+import { Box } from '@mui/material';
+
+function TagList({ children }) {
+  const tagsHead = [];
+  const tagsTail = [];
+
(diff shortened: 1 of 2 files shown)


dadiorchen:
@RubenSmn I will add this, but in my feeling, I think it is a little bit overkill, I think another consideration is the humen understanding factor, the code should give the developer a limited mindset rather than give them extra consideration/knowlege.

RubenSmn:
@dadiorchen I will add some comments on the `TagList` to help understand how to component works. Should I also refactor the other id pages?

PR #912, opened 2022-09-08
feat: show timeline only on global context

### LINKED ISSUE #911: The timeline should only be available in a `global` context
Currently, we just support timeline for a global map, so if the path is in a context like: planter, org, wallet, then we should hide the timeline to avoid confusion. like this path: http://localhost:3000/organizations/178/trees/762681

RubenSmn:
# Description

Fixes #911

## Type of change

# Checklist:

### DIFF
--- a/src/components/Timeline.js
+++ b/src/components/Timeline.js
@@ -9,6 +9,7 @@ import { useState, useEffect } from 'react';
 import { useMapContext } from 'mapContext';
 import { useMobile } from '../hooks/globalHooks';
 import TimeIcon from '../images/icons/time.svg';
+import { nextPathBaseDecode } from '../models/utils';
 
 function CustomInput({ label, inputRef, inputProps, InputProps }) {
   return (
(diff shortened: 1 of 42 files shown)


RubenSmn:
I will fix the checks in a bit

dadiorchen:
@RubenSmn is this pr ready for merge?

RubenSmn:
@dadiorchen yes it only shows on the `top` page now

dadiorchen:
@RubenSmn it should be alse visible on home/index page

RubenSmn:
I will fix that!

src/components/Timeline.js
```diff
@@ -68,7 +68,7 @@ function Timeline() {
     end: Date.now(),
   });
   const { map } = useMapContext();
-  const isGlobalPage = !!router.asPath.match(/\/top/g);
+  const isGlobalPage = !!router.pathname.match(/^(?:\/top|\/)$/g);
```
dadiorchen, line 71: @RubenSmn we need to do some extra work on these cases, the path on other envs will be a bit different with local, the root of next.js will be: /web-map-beta/beta as the base path, os we will fail on your code above on other envs, please check my code here to handle it: https://github.com/Greenstand/treetracker-web-map-client/blob/752f85020c90473c4fd56940edb6f6ef82ca58b6/src/components/App.js#L76-L90

RubenSmn: Thanks for the extra info!

dadiorchen:
@RubenSmn sorry for the delay, can you take a look at the conflict?

RubenSmn:
I will!

RubenSmn:
@dadiorchen fixed the conflicts

PR #930, opened 2022-09-09
feat: better way to handle errors

### LINKED ISSUE #821: A better way to handle 404 and 500
We handle the error like this:

``
export async function getServerSideProps({ params }) {
  const id = params.walletid;
  try {
    const [wallet, species, tokens, tokenRegionCount] = await Promise.all([
      getWalletById(id),
      getSpeciesByWalletId(id),
      (async () => {
        // Todo write a filter api that only returns totalNo.of tokens under a certain wallet
        const data = await requestAPI(`/tokens?wallet=${id}`);
        return data;
      })(),
      (async () => {
        // return total no.trees/tokens per country
        const data = await requestAPI(`/wallets/${id}/token-region-count`);
        return data.walletStatistics;
      })(),
    ]);
    return {
      props: {
        wallet,
        species: species.species,
        tokens,
        tokenRegionCount,
      },
    };
  } catch (e) {
    log.warn("e:", e);
    return { notFound: true };
  }
}
```

This is good to avoid the app crashing, but it is not accurate, I experience an error in the try/catch, it's just simple wrong code, but the page is showing 404 not found, maybe we need to distinguish what happened, if the data fetching returns 404, then we can display this, but if something else, we need to jump to 500 (something went wrong) page.

ar563:
# Description

- better way to handle errors
- new page: 500

Fixes #821

## Type of change

# How Has This Been Tested?

- [ ] Cypress integration
- [ ] Cypress component tests
- [x] Jest unit tests

# Checklist:


dadiorchen:
I think this is a good appraoch.

RubenSmn (approved):

RubenSmn:
I think this is an good approach aswell! We can probably implement this on every page right?

ar563:
> I think this is an good approach aswell! We can probably implement this on every page right?

done

Issue #498, opened 2022-03-02
Visual test with Cypress


dadiorchen:
Do some research for visual testing with Cypress, the goal is to detect significant UI change with the new PR, so we can check/decide if it is a bug or not.

https://docs.cypress.io/guides/tooling/visual-testing#Applitools

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.
- More issues in this project: https://github.com/orgs/Greenstand/projects/26

RubenSmn:
I will do some research on this @dadiorchen

RubenSmn:
They recommend using Cypress 10 https://applitools.com/tutorials/quickstart/web/cypress#step-2-getting-your-example-project
We're currently using Cypress 8

PR #937, opened 2022-09-11
chore: add before/after screenshot tempate

RubenSmn:
# Description
Add before/after template for screenshots for a more efficient workflow
- Easier to compare for UI changes

## Type of change


dadiorchen:
:D great, yes, it's useful. and I was thinking if we can do the visual test by cypress, then we can relieve people on doing this, because the cypress will automatically list the difference of the result in UI.

dadiorchen:
As metioned here: https://github.com/Greenstand/treetracker-web-map-client/issues/498

PR #951, opened 2022-09-14
fix: 500 page showing in production env

### LINKED ISSUE #942: The correct 500 page is flashing and then dispear
To reproduce, open this page: https://dev-k8s.treetracker.org/web-map-beta/demo/trees/top
There was a correct 500 page show up, but immediately replaced by this :

Image

RubenSmn:
# Description

Fixes #942

## Type of change

# Checklist:

### DIFF
--- a/src/pages/organizations/[organizationid].js
+++ b/src/pages/organizations/[organizationid].js
@@ -521,6 +521,11 @@ export async function getServerSideProps({ params }) {
   } catch (e) {
     log.warn('organizations page:', e);
     if (e.response?.status === 404) return { notFound: true };
-    throw e;
+    return {
+      redirect: {
+        destination: '/500',
+        permanent: false,
+      },
+    };
   }
 }
--- a/src/pages/organizations/[organizationid].js
+++ b/src/pages/organizations/[organizationid].js
@@ -505,27 +505,16 @@ export default function Organization(props) {
   );
 }
 
-export async function getServerSideProps({ params }) {
+export const getServerSideProps = utils.wrapper(async ({ params }) => {
   const id = params.organizationid;
-  try {
-    const organization = await getOrganizationById(id);
-    const orgLinks = await getOrgLinks(organization.links);
-    return {
-      props: {
-        organization: {
-          ...organization,
-          ...orgLinks,
-        },
+  const organization = await getOrganizationById(id);
+  const orgLinks = await getOrgLinks(organization.links);
+  return {
+    props: {
+      organization: {
+        ...organization,
+        ...orgLinks,
       },
-    };
-  } catch (e) {
-    log.warn('organizations page:', e);
-    if (e.response?.status === 404) return { notFound: true };
-    return {
-      redirect: {
-        destination: '/500',
-        permanent: false,
-      },
-    };
-  }
-}
+    },
+  };
+});
--- a/src/models/utils.js
+++ b/src/models/utils.js
@@ -223,6 +223,18 @@ function nextPathBaseDecode(path, base) {
   return path.replace(base, '');
 }
 
+const wrapper = (callback) => (params) =>
+  callback(params).catch((e) => {
+    log.warn('error retrieving server props:', e);
+    if (e.response?.status === 404) return { notFound: true };
+    return {
+      redirect: {
+        destination: '/500',
+        permanent: false,
+      },
+    };
+  });
+
 export {
   hideLastName,
   parseDomain,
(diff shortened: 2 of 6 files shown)


src/pages/organizations/[organizationid].js
```diff
@@ -521,6 +521,11 @@ export async function getServerSideProps({ params }) {
   } catch (e) {
     log.warn('organizations page:', e);
     if (e.response?.status === 404) return { notFound: true };
-    throw e;
+    return {
+      redirect: {
+        destination: '/500',
+        permanent: false,
+      },
+    };
```
Comment (dadiorchen) [inline src/pages/organizations/[organizationid].js:529]: Do you think we can use a function or so to reduce the boilerplate, like:

```
export async const getServersideProps = serverSiderPropsWrapper(() => {
...
})

function serverSiderPropsWrapper(callback){
   try{
       return callback();
    }catch(
....
}
```

Comment (dadiorchen) [reply src/pages/organizations/[organizationid].js:529]: I'm not sure this is possible, but seems the code is a bit duplicated.

Comment (RubenSmn) [reply src/pages/organizations/[organizationid].js:529]: Yes I was just think about this, we could turn this into a function! But what happens if for example in the tree page, we have the tree. But the organizations fails. Then we still get the 500 page, even with the previous way of handling the error

Comment (dadiorchen) [reply src/pages/organizations/[organizationid].js:529]: I don't get it, what's `the organizations fails` ? 

I think we just extract the code and put into a function, so we can reuse it?

Comment (dadiorchen) [reply src/pages/organizations/[organizationid].js:529]: Isn't the catch part (the code and logic) the same for every page?

Comment (RubenSmn) [reply src/pages/organizations/[organizationid].js:529]: For example here:
https://github.com/Greenstand/treetracker-web-map-client/blob/beta/src/pages/trees/%5Btreeid%5D.js#L679-L689

What if the tree request is fine but the planter or organization request fails. Then the user will be redirected to the 500 page even if we have all the data from the tree itself.

Comment (dadiorchen) [reply src/pages/organizations/[organizationid].js:529]: But I don't think this is relevant with the extracting-function-topic, right?
Back to your example, I think it's fine to return 500 if any requests failed in the page

Comment (dadiorchen) [reply src/pages/organizations/[organizationid].js:529]: We'd better arrange a meeting for this if we can not agree with each other?

Comment (RubenSmn) [reply src/pages/organizations/[organizationid].js:529]: If it is fine to return to the 500 if any request fails then its not a problem. I will refacter this to some wrapper for this!

RubenSmn:
I am having a bit of trouble implementing the wrapper function, so far I have this

```js
// [treeid].js

import * as utils from "../utils";

export const getServerSideProps = ({ params }) =>
  utils.serverSidePropsWrapper(async () => {
    const { treeid } = params;
    const tree = await getTreeById(treeid);
    const { planter_id, planting_organization_id } = tree;
    const planter = await getPlanterById(planter_id);
    let organization = null;
    if (planting_organization_id) {
      log.warn('load org from planting_orgniazation_id');
      organization = await getOrganizationById(planting_organization_id);
    } else if (planter.organzation_id) {
      log.warn('load org from planter. organization_id');
      organization = await getOrganizationById(planter.organization_id);
    }

    return {
      props: {
        tree,
        planter,
        organization,
      },
    };
  });
```

```js
// utils.js

const serverSidePropsWrapper = (callback) => {
  try {
    return callback();
  } catch (e) {
    log.warn('error retrieving server props:', e);
    if (e.response?.status === 404) return { notFound: true };
    return {
      redirect: {
        destination: '/500',
        permanent: false,
      },
    };
  }
};
```

This way I am still getting the error mentioned in #942 
While handling the errors the way in this commit https://github.com/Greenstand/treetracker-web-map-client/pull/951/commits/07a914fab5eaeccb20b7bd6d72bdd9acd9dd5d42 everything works just fine

Am I missing something?

dadiorchen:
I think you need to `await` at this line?
``
    return callback(); 
```

Or you can:

```
callback()
.catch(e => {
...
})
```

**Comment by dadiorchen** [comment]
To make the caller simpler, maybe we can do this:

```
export const getServerSideProps = utils.wrapper(params => {
  ...
})
```

```
const wrapper = callback => {
  return params => {
    return callback(params)
                .catch(e => {
                   return {error:500}
                 })
}

RubenSmn:
Thank you if forgot the `await`!

dadiorchen:
@RubenSmn sorry for causing this conflict, can you resolve it?

PR #1018, opened 2022-09-27
fix: broken layout for longer names

### LINKED ISSUE #992: Planter page: broken layout
https://prod-k8s.treetracker.org/web-map-beta/demo/planters/3939?keyword=undefined

The species component get distort because of long text

Image

khalatevarun:
# Description

Fixes #992 

**Problem**: The layout for card was broken for longer names.

**Solution**: Add wordWrap and fixed width property to limit the name up to particular length and shift the rest to next line.

_NOTE: increased padding for inner card content, the height of the card is slightly increased._

## Type of change

## Screenshots

| 191858432-41a0f344-c4d4-4c7c-8ba6-3a7e84c47e54|SCR-20220927-way|

# How Has This Been Tested?

- [ ] Cypress integration
- [ ] Cypress component tests
- [ ] Jest unit tests

# Checklist:

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/src/components/TreeSpeciesCard.js
+++ b/src/components/TreeSpeciesCard.js
@@ -38,7 +38,6 @@ function TreeSpeciesCard(props) {
           sx={{
             wordWrap: 'break-word',
             width: '65%',
-            padding: '8px 0',
           }}
         >
           <Typography
@@ -78,7 +77,7 @@ function TreeSpeciesCard(props) {
           sx={{
             background: (t) => t.palette.background.paperDark,
             color: (t) => t.palette.text.primaryReverse,
-            padding: (t) => [t.spacing(4, 5), t.spacing(6, 8)],
+            padding: (t) => [t.spacing(5), t.spacing(6, 8)],
             display: 'flex',
             flexDirection: 'column',
             alignItems: 'center',
--- a/src/components/TreeSpeciesCard.js
+++ b/src/components/TreeSpeciesCard.js
@@ -38,6 +38,7 @@ function TreeSpeciesCard(props) {
           sx={{
             wordWrap: 'break-word',
             width: '65%',
+            padding: '8px 0',
           }}
         >
           <Typography
@@ -77,7 +78,7 @@ function TreeSpeciesCard(props) {
           sx={{
             background: (t) => t.palette.background.paperDark,
             color: (t) => t.palette.text.primaryReverse,
-            padding: (t) => [t.spacing(3, 5), t.spacing(6, 8)],
+            padding: (t) => [t.spacing(4, 5), t.spacing(6, 8)],
             display: 'flex',
             flexDirection: 'column',
             alignItems: 'center',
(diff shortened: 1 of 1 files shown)


src/components/TreeSpeciesCard.js
```diff
@@ -70,7 +77,7 @@ function TreeSpeciesCard(props) {
           sx={{
             background: (t) => t.palette.background.paperDark,
             color: (t) => t.palette.text.primaryReverse,
-            padding: (t) => [t.spacing(3, 5), t.spacing(6, 8)],
+            padding: (t) => [t.spacing(4.5, 6.5), t.spacing(6, 8)],
```
RubenSmn, line 80: Personally I do not see the need of a padding increase, what is your opinion @jeremydthomas ?

RubenSmn: I see the issue now, when wordwrapping the text can become a little close to the border

src/components/TreeSpeciesCard.js
```diff
@@ -32,7 +32,14 @@ function TreeSpeciesCard(props) {
         justifyContent="space-between"
         alignItems="center"
       >
-        <Grid item ml={6}>
+        <Grid
+          item
+          ml={6}
+          sx={{
```
RubenSmn, line 38: @khalatevarun could you add some vertical padding here and undo the change other (count element) padding

RubenSmn (changes requested):

src/components/TreeSpeciesCard.js
jeremydthomas: I haven't actually looked at it more than a glance to be honest. What I did see seemed better than what I did. I was working on something else tonight, also about to go home here in a bit.

src/components/TreeSpeciesCard.js
khalatevarun: I tried doing the same initally but as a consequence of that, there is additional white space created on top and bottom of count element.
Please refer to image attached. 
SCR-20220928-s54

khalatevarun: I think adding padding for the count element is would be better as it would result only slight increase in the list height.

RubenSmn: We could fix this by adding `align-self: stretch` to the count box.

RubenSmn: If we don't add the padding to the description side of the box the spacing will not look the same for all cards
image

khalatevarun: makes sense, but it would result in cards of different heights since some name will take 2 lines and adding vertical padding will further increase vertical height thus resulting in inconsistency?

Can surely make changes suggested by you if the above mentioned concern is not a big deal. 
Let me know if I should undo my changes and add vertical padding.

RubenSmn: I think its fine if height increases when we have more than 1 line.

khalatevarun: okay, will update changes

RubenSmn:
@dadiorchen this was not ready yet, we are still having some spacing issues if the specie name is really long or if the device width is small

image

dadiorchen:
Oh, what's should I do now? revert?

khalatevarun:
> 

I have created another PR with updated changes.

PR #1020, opened 2022-09-27
feat: on demand revalidation route

### LINKED ISSUE #998: Research: could we manually trigger the cache update for next.js
Currently, we are going to use the next.js's cache to speed up the performance, and set an expiration time for that. Can we find a way to manually trigger some page's cache update so we can get the newest data immediately.

This is to bring convenience for our operator, usually, after doing some data management, they want to check the result on the map immediately, but with the cache of next.js, they need to wait.

okdv:
# Description

API route for forcing revalidation of nextjs cache

User must have .env.local setup and it should contain variable REVALIDATION_SECRET, which will be supplied in api call for extra security 

Can then revalidate paths with following url (local example)
GET http://localhost:3000/web-map-beta/demo/api/revalidate?secret=TOKEN&path=/top
401 if token is bad
404 if path is bad
500 if error with revalidation function
200 if success

Will add applicable README changes in this PR, once we confirm this is built as desired

Will also add a test for this change, again once confirm its built as desired

Fixes #998 (discovery thread with @dadiorchen )

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

# How Has This Been Tested?

- [X] Cypress integration
- [X] Cypress component tests

# Checklist:

### DIFF
--- a/src/pages/api/revalidate.js
+++ b/src/pages/api/revalidate.js
@@ -1,0 +1,17 @@
+export default async function handler({ query }, res) {
+  const { secret, path } = query;
+  if (secret !== process.env.REVALIDATION_SECRET) {
+    return res.status(401).json({ message: 'Invalid token' });
+  }
+
+  if (!path) {
+    return res.status(401).json({ message: 'Invalid path' });
+  }
+
+  try {
+    await res.revalidate(`${process.env.NEXT_PUBLIC_BASE ?? ''}${path}`);
+    return res.json({ revalidated: true });
+  } catch (err) {
+    return res.status(500).send('Error revalidating');
+  }
+}


okdv:
meant to open as draft, changed it after the fact

dadiorchen:
Oh, this is amazing, thanks for helping to consider the security problem.

okdv:
@dadiorchen i think this takes care of the cache revalidation IIRC, i just never finished a small bug and put it out of draft. I'll take care of that

okdv:
fixed the little bug i mentioned but had forgotten about, submitting for review now 

might be a good idea to add a bit into the README about this as well, thoughts on if that is necessary @dadiorchen ?

Issue #976, opened 2022-09-21
Improve the UX for `next` tree


dadiorchen:
Currently, if we click next tree button, the next.js will load the data in ajax-like way, this is good to reduce the network traffic, but if the net condition is slow, the experience is not so good, the user don't know what's happening after clicking the button, see video below, maybe we should find a way to indicate the progress of loading, like the process bar on the top of the page or so?

https://dadior.s3.ap-northeast-1.amazonaws.com/Screen%20Recording%202022-09-21%20at%204.12.43%20PM.mp4

To reproduce: https://prod-k8s.treetracker.org/web-map-beta/demo/wallets/1f2a0862-66d1-4b42-8216-5a5cb9c6eca5/tokens?tree_id=95602

dadiorchen:
Acutally, I think this is a common case on our site, for example, on the top page, click organization  Bôndy, and it takes dozens of seconds to open new page, but before that, the UI is doing nothing to let the user know what's happening.

okdv:
Definitely seems to be a few approaches, we can just inject the loading indicator at the App level so it works for all page loads. I'd just say add a delay before we show the indicator so it doesnt try to pop up even when the page loads quickly. 

Could keep it really simple and use something like [nprogress](https://www.npmjs.com/package/nprogress)?

dadiorchen:
@okdv good, if the mui component works for this, https://mui.com/material-ui/react-progress/#main-content I prefer to use it directly, but if it doesn't I am good to choose another lib, 

and also I found some discussion related: https://stackoverflow.com/questions/55624695/loading-screen-on-next-js-page-transition

PR #1045, opened 2022-09-30
feat: loading spinner

### LINKED ISSUE #1039: Display progress block when loading new page
As the video showing, it is a bit confusing for user when we click next tree,  the old id and picture is still there and replaced by new ones when loaded, we should remove old content and replace with loading block (like this https://mui.com/material-ui/react-skeleton/#main-content) and display new content when ready.

https://dadior.s3.ap-northeast-1.amazonaws.com/Screen%20Recording%202022-09-30%20at%205.51.34%20PM.mp4?

Note, this issue is related to #976

okdv:
# Description

Loading spinner for side panel/drawer on route change. chose to design in a way that it is an overlay with the previous children still underneath, visible behind semi transparent background with simple spinner in the center
At _app level trigger functions to set loading prop based on router events
At Layout levels, render new spinner overlay component if loading is true and 500ms have passed (this way it does not appear for quickly loaded content)

Fixes #1039 #976 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Screenshots

After:
image

# How Has This Been Tested?

- [X] Cypress integration
- [X] Cypress component tests

# Checklist:

- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -2,9 +2,9 @@ import ExpandLessIcon from '@mui/icons-material/ExpandLess';
 import { Box, Paper } from '@mui/material';
 import log from 'loglevel';
 import React from 'react';
+import SpinnerOverlay from "./SpinnerOverlay";
 
-export default function Drawer(props) {
-  const { children } = props;
+export default function Drawer({ children, loading }) {
   const rootRef = React.useRef(null);
   const buttonRef = React.useRef(null);
   const contentRef = React.useRef(null);
@@ -158,6 +158,13 @@ export default function Drawer(props) {
     }
   }, []);
 
+  const dimensions = {
+    width: '100%',
+    height: '100vh',
+    top: 0,
+    left: 0,
+  };
+
   React.useEffect(() => {
     log.warn('mount listener...');
     rootRef.current.addEventListener('touchstart', handleTouchStart);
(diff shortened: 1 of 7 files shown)


dadiorchen:
@okdv this is really cool!

src/components/Drawer.js
```diff
@@ -158,6 +158,13 @@ export default function Drawer(props) {
     }
   }, []);
 
+  const dimensions = {
+    width: '100%',
+    height: '100vh',
+    top: 0,
+    left: 0,
+  };
+
```
dadiorchen, line 167: @okdv do you have to add this to drawer? is it possible to put it somewhere else? The drawer is pretty fragile I feel, I prefer to sacrify something to keep it simple

dadiorchen: Oh, I guess the code in drawer is to solve issue #1039 , could you slit it to two PR for these two respectively?

dadiorchen:
@okdv can you split this to two PR? for the two issues separately?

src/components/Drawer.js
okdv: @dadiorchen didn't actually test it without those settings, since I had just grabbed them from the existing style of the drawer. But I can look into seeing if something if its necessary?

dadiorchen:
@okdv you work did help solving the problem, yes, I merged your pr and tested, I found some problem, so I decided to revert your pr and manually copy your code into the repo, and made some change:

1. remove the setTimeout, I know the intention, but it brings problem: when you jump back to a page and it loaded quickly, then, the spin animation will stay on the top forever, I think this is because the `complete` even get executed before the setTimeout function, so I removed it.
2. I'm afraid to touch drawer, and also, the drawer solution faces problem in this case: if the user slaps the drawer to the bottom, then we can not see the loading progress, so I change to using the simple linear progress on the top.

anyway, thanks a lot for help us on this!

dadiorchen:
@okdv I hope you are well!

PR #1052, opened 2022-10-02
fix(featured-card): update spacings conform design

RubenSmn:
# Description

- update spacings to better align with the [figma design](https://www.figma.com/file/XdYFdjlsHvxehlrkPVYq0l/Greenstand-Webmap?node-id=8229%3A33574)

## Type of change

## Screenshots

|image|image|

# Checklist:

### DIFF
--- a/src/components/FeaturedTreesSlider/index.js
+++ b/src/components/FeaturedTreesSlider/index.js
@@ -108,12 +108,19 @@ function FeaturedTreesSlider({ trees, size = null, isMobile, link }) {
                   minWidth: [144, 208],
                 }}
               />
-              <CardContent>
+              <CardContent
+                sx={{
+                  p: 0,
+                  ':last-child': {
+                    p: 0,
+                  },
+                  mt: 5,
+                }}
+              >
                 <Typography
                   variant="h6"
                   sx={{
                     fontSize: '20px',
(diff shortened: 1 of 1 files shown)


RubenSmn:
We might want to add some more `margin` on the text content, but I do think it is better if the left side of the text aligns with the left side of the image.

dadiorchen:
@RubenSmn sorry, I want to confirm, you think align with the figma design is better, with fewer space/margin upside and downside, right?

RubenSmn:
@dadiorchen to be honest, no. Like I mentioned in my previous comment I do think the text should align with the image (horizontal spacing). As for the vertical spacing it could be reduced just a little

Quick question, the figma design is a guideline and we can tweak/change things if we think they're better?

dadiorchen:
I think the principle of this is: in terms of UI/appearance, we should respect the design file, but if the UI doesn't fit our functionality, we can change the design, for example, for the feature planter, and organization slider below, to make the UI neater, the UI designer ignored the name of the org/planter, but in actual case, it's a high possibility that we don't have the logo of the organization, so maybe display the name of the org is a better solution, in this case, we should change the UI.
If it's about the appearance of the UI, we should follow the desgin as much as possible.

PR #1086, opened 2022-10-10
feat: render feature section only if data exists

### LINKED ISSUE #1072: Hide the featured section if it is empty
On the `top` page, it is possible that we won't display some of the featured sections:

Image

So add logic that if the featured array from API is empty,then hide this section.

khalatevarun:
# Description

Fixes #1072 

## Type of change

## Screenshots

# How Has This Been Tested?

- [ ] Cypress integration
- [ ] Cypress component tests
- [ ] Jest unit tests

# Checklist:

### DIFF
--- a/src/pages/top.js
+++ b/src/pages/top.js
@@ -132,49 +132,67 @@ function Top(props) {
             }}
           />
         </Box>
-        <Box
-          sx={{
-            mt: 8,
-          }}
-        >
-          <Typography variant="h4">Featured trees this week</Typography>
-        </Box>
-        {false && ( // going to be replaced by search filter component
-          <Box sx={{ display: 'flex', justifyContent: 'flex-end' }}>
-            <Filter onFilter={handleFilter} />
-          </Box>
+        {trees?.length > 0 && (
+          <>
+            <Box
+              sx={{
+                mt: 8,
+              }}
+            >
+              <Typography variant="h4">Featured trees this week</Typography>
+            </Box>
+            {false && ( // going to be replaced by search filter component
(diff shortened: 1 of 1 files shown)


khalatevarun:
@RubenSmn @dadiorchen please review.

RubenSmn (commented):
There are a lot of places that the `Featured` slider is being used.

I think we should handle the check "inline". Just pass in the trees or the organizations and let the component handle the check, return the slider if there is data else return null.

This does result in a new issue; the title will still be visible and will need the check anyway. I would like to suggest that we give the `Featured` slider a `title` prop and we display the title that way since the components/style for the title all seems to be the same.

@dadiorchen what do you think?

khalatevarun:
@RubenSmn but if the trees/organisations data is empty then we will have to hide the title, so I think that the condition to hid the title depends on if the tree/organisation data is empty or not which is why we will have to keep it outside both the components.

dadiorchen:
@khalatevarun I think the code is fine.

Yes, there is way to test it, for this `edge case` you can create a new integration test and MOCK the data to pass EMPTY list into the page (the `top.cy.js`)

dadiorchen:
Would you like to try it? I mean, build the test to cover it?

khalatevarun:
Sure @dadiorchen i would like to give it a try. Allow me some 2-3 days of time.

khalatevarun:
@dadiorchen can we go ahead with this, couldn't find time for creating test case. Will definitely pick up later.

dadiorchen:
@khalatevarun that's totally fine! Let's build test case for future case when it's getting really necessary!

PR #1135, opened 2022-10-22
fix: Make slider button visible

### LINKED ISSUE #1057: The forward and back buttons on the tree slider / sliders are hard to see.
As you can see in the picture there is an arrow forward, and when you go forward there is an arrow that goes backwards that is not in the picture. If I didnt know that was there, I dont think I would see it, so I doubt website visitors would see it either. 

Try to make this more visible so that the visitors know its there.

Inkedimage

ojafero:
# Description

Fixes #1057

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Screenshots

| before | after |

# Checklist:


dadiorchen:
@ojafero thanks, this is beautiful

PR #1226, opened 2022-12-02
chore: rename 'Learn more' button text

ojafero:
# Description

Rename "Learn more" to a descriptive text for SEO.

# Checklist:

### DIFF
--- a/src/components/Home.js
+++ b/src/components/Home.js
@@ -138,7 +138,7 @@ export default function Home(props) {
                   },
                 }}
               >
-                Learn more
+                Read about Treetracker
               </Box>
             </Link>
           </Button>


jeremydthomas:
@ojafero could you throw up a picture of your work? Would help reviewers save some time to look at what you've done without loading up your code themselves, especially for a few lines of code.

src/components/Home.js
```diff
@@ -138,7 +138,7 @@ export default function Home(props) {
                   },
                 }}
               >
-                Learn more
+                Read about Treetracker
```
jeremydthomas, line 141: I think "Read More About Treetracker" or "Read More About The Treetracker System" would be a bit better. Its also better to capitalize the first letter of each word when it comes to seo. What do you think of those phrases @dadiorchen? This is a great step in the right direction though!

jeremydthomas (changes requested):

jeremydthomas:
Oh, I just noticed how old this PR is 😆

dadiorchen:
Because this changed our UX, I will close this PR

PR #1233, opened 2022-12-02
chore: update organization card text

ojafero:
# Description

Ass. is an unclear abbreviation. Better to include all the letters.


dadiorchen:
Yes, good for SEO too

PR #1238, opened 2022-12-02
chore: replace console logs

ojafero:
# Description
Replace console logs with loglevel.

# Checklist:


dadiorchen:
@ojafero thanks for doing this, loglevel is a good way to build a robust enterprise level of application

PR #1239, opened 2022-12-03
fix: kebab case warning

ojafero:
# Checklist:

### DIFF
--- a/src/components/FeaturedPlantersSlider/index.js
+++ b/src/components/FeaturedPlantersSlider/index.js
@@ -152,8 +152,8 @@ function FeaturedPlantersSlider({
                   textAlign: 'center',
                   wordBreak: 'break-word',
                   display: '-webkit-box',
-                  '-webkit-line-clamp': '2',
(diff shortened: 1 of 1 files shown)


src/components/FeaturedPlantersSlider/index.js
```diff
@@ -152,8 +152,8 @@ function FeaturedPlantersSlider({
                   textAlign: 'center',
                   wordBreak: 'break-word',
                   display: '-webkit-box',
-                  '-webkit-line-clamp': '2',
-                  '-webkit-box-orient': 'vertical',
+                  WebkitLineClamp: '2',
+                  WebkitBoxOrient: 'vertical',
```
dadiorchen, line 156: shouldn't be webkitLineClamp, I feel camel case is the convention on the CSS name in react?

ojafero: @dadiorchen This is a special case. I checked Devtools and these css properties are applied. More info: https://github.com/emotion-js/emotion/issues/2036#issuecomment-705159259

PR #1241, opened 2022-12-03
fix: DOM warning

ojafero:
# Description

Remove div nested in p.

## Type of change

# Checklist:

### DIFF
--- a/src/pages/wallets/[walletid].js
+++ b/src/pages/wallets/[walletid].js
@@ -353,13 +353,13 @@ export default function Wallet(props) {
         <Typography sx={{ mt: [2.5, 5] }} variant="h4">
           About the Wallet
         </Typography>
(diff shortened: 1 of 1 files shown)


src/pages/wallets/[walletid].js
```diff
@@ -353,13 +353,13 @@ export default function Wallet(props) {
...
-            }}
-          />
-        </Typography>
+        <Typography
+          sx={{ mt: [2.5, 5] }}
+          variant="body2"
+          dangerouslySetInnerHTML={{
+            __html: marked.parse(wallet.about || 'NO DATA YET'),
+          }}
+         />
```
Comment (dadiorchen) [inline src/pages/wallets/[walletid].js:362]: @ojafero did you test it with data of the wallet.about? I remember that's on purpose with that div, otherwise the data fails to display?

Comment (ojafero) [reply src/pages/wallets/[walletid].js:362]: @dadiorchen Yes, it works.

PR #1245, opened 2022-12-05
fix: update card shadow

ojafero:
# Description

Update FeaturedPlantersSlider shadow to match Figma design.

## Screenshots

|  | |

# Checklist:


dadiorchen:
Elevation is not just about shadow, it takes something else into consideration, to achieve the goal of material design's layer principle, and also to fit dark theme, and because we are now considering making the web map a theme-customizable app, we need to find a better way to change the shadow, maybe do it in the theme template. @RubenSmn

RubenSmn:
@dadiorchen that is a good addition for the cwm.

PR #1251, opened 2022-12-05
fix: alignment issue

ojafero:
# Description

Align TreeTags on the wallet page.

## Type of change

## Screenshots

| not aligned | aligned |

# Checklist:


dadiorchen:
@ojafero because we are using treetag in other pages, so I feel this will change the layout on those cases, the original design of the tag is to set the width by content, so we got a pile of tags filling up the space.
And also, for this case, I think there is no need to display the whole uuid, we can use the uuid component to shorten them, just like you did for another case.

Issue #830, opened 2022-08-19
Link component can not use the class to set style


dadiorchen:
For this commit: 

https://github.com/Greenstand/treetracker-web-map-client/pull/829/commits/c4c97cac4166b4d0372e02e88572ea25fac675ba

In the code, I have to set the inline style to specify CSS property, I can not use the className by mui to set it, inline style is not the best solution, so we need to fix the problem.

I wanted to add the style in the class.link, but it doesn't work. to reproduce this problem, you can try to add some inline style into the `a` and check the result on the browser, then, if you move it to the class.link object, it will not work.

I think the reason is that when we render the `a` it is happening on the server side, so maybe the random class name is wrong when we run it on the client side?

keyy123:
Is this issue still open, @dadiorchen? I don't mind adding it to my queue of issues to tackle so I save time not looking for more issues. 😁👌

Since this is my first time working with the project's client side, Is there any set up that I need to do before I can handle this issue?

dadiorchen:
@keyy123 are you on our slack community? if not , please leave your email so I can invite you, for the set up, I recommend you go through the readme, there are some guide and heads up

keyy123:
@dadiorchen  I am already in the slack group for web-map! 😁👌

dadiorchen:
I think we should be more careful on changing the basic class and style for this core common component, so I think we'd better discard the idea of styling the link by class but directly changing inline style instead.

PR #1254, opened 2022-12-06
Challenge/use class styles for link component

keyy123:
# Description

Apply styling for the Link component via class names. It also includes comments that leave warnings about inline-theme style override using the MUI sx prop and references to theme styling in code.  

Fixes # 830

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Screenshots

|                                                             |

[Before and After](https://www.loom.com/share/4a2fa769ccd44e9d8c12ffe959fd8748 )

# How Has This Been Tested?

- [X] Cypress integration
- [X] Cypress component tests
- [X] Jest unit tests

# Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/components/Home.js
+++ b/src/components/Home.js
@@ -1,3 +1,4 @@
+import { Brightness1 } from '@mui/icons-material';
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import Typography from '@mui/material/Typography';
@@ -130,13 +131,18 @@ export default function Home(props) {
               target="_blank"
               href="https://greenstand.org/treetracker/web-map"
             >
+              {/* Box Component sx attribute overrides the 
+                style within Link Component so be careful */}
+
               <Box
-                sx={{
-                  color: 'white',
-                  '.MuiBox-root&:hover': {
-                    color: '#adadad',
-                  },
-                }}
+                sx={
+                  {
+                    // color: 'white',
+                    // '.MuiBox-root&:hover': {
+                    //   color: '#adadad',
+                    // },
+                  }
+                }
               >
                 Learn more
               </Box>
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -9,8 +9,10 @@ import React from 'react';
 // generate material-ui style hook
 const useStyles = makeStyles((theme) => ({
   link: {
-    color: theme.palette.text.primary,
+    color: theme.palette.text.primaryReverse,
     '& :hover': {
+      // You can feel the team's established theme styles in themeContext.js
+      // Lines 12 - 327
       color: theme.palette.text.primary,
     },
   },
@@ -19,8 +21,9 @@ const useStyles = makeStyles((theme) => ({
 const NextComposed = React.forwardRef((props, ref) => {
   const { as, href, ...other } = props;
   const classes = useStyles();
+  console.log(props.children);
   return (
-    <NextLink href={href} as={as}>
+    <NextLink href={href} as={as} passHref>
       <a
         style={{ display: 'block' }}
         ref={ref}
--- a/src/components/playground/FontAddInput.cy.js
+++ b/src/components/playground/FontAddInput.cy.js
@@ -33,6 +33,7 @@ describe('Toggle Theme Mode', () => {
     );
     cy.get('.MuiAccordionSummary-content').click();
     cy.get('.MuiFormControl-root').as('formControl');
+    // using a timeout prevents cypress-no-arbitrary-wait error
     cy.get('@formControl')
       .eq(0)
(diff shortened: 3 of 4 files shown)


keyy123:
My only question is should I remove the sx attribute from the `<box>` component?

src/components/Home.js
```diff
@@ -130,13 +131,18 @@ export default function Home(props) {
...
+                style within Link Component so be careful */}
+
               <Box
-                sx={{
-                  color: 'white',
-                  '.MuiBox-root&:hover': {
-                    color: '#adadad',
-                  },
-                }}
+                sx={
```
RubenSmn, line 138: I think these are fine so no need to change them here. But you did help me find a problem!

RubenSmn (changes requested):
I left some notes. The key problem here is that we set an *inline* style `display: 'block'` we need to find a solution for that problem.

RubenSmn:
@dadiorchen I just tried to add the style to the class like so
```js
  link: {
    display: 'block',
    color: theme.palette.text.primary,
    '& :hover': {
      color: theme.palette.text.primary,
    },
  },
```
And build it locally and it seems to work fine. Could you point out were this is happening?
All the `a` tags had this `class`
```js
.jss1 {
  color: #eee;
  display: block;
}
```

src/components/Link.js
```diff
@@ -9,8 +9,10 @@ import React from 'react';
 // generate material-ui style hook
 const useStyles = makeStyles((theme) => ({
   link: {
-    color: theme.palette.text.primary,
+    color: theme.palette.text.primaryReverse,
```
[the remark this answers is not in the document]
keyy123: Should I revert style to primary?

src/components/playground/FontAddInput.cy.js
```diff
@@ -32,7 +32,12 @@ describe('Toggle Theme Mode', () => {
       </PlaygroundProvider>,
     );
     cy.get('.MuiAccordionSummary-content').click();
-    cy.get('.MuiFormControl-root').eq(0).type('void void vpid{enter}');
-    cy.get('.MuiFormControl-root').eq(0).contains('Font could not be loaded');
+    cy.get('.MuiFormControl-root').as('formControl');
```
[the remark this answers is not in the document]
keyy123: Yes @RubenSmn 😁👌

src/components/Home.js
keyy123: I will revert back changes in Home.js then 🙂.

keyy123:
@RubenSmn what component had that class?

keyy123:
@RubenSmn, MakeStyles has a thing with it classes. In dev environments, class name is "makestyles-${className}-${#}" (i.e., classes.link === makestyles-link-1) but in production, the classname is "jss${#}" (classes.link === jss1). In this case, every `<Link>` component in the app (~20 in web-map-client) will have a jss1 class name in prod b/c we attached make styles link class to the sub-component within the link component. 

TLDR: The styles that you see come from MakeStyles  - it's called jss1 in prod and makestyles-link-1 in dev

RubenSmn:
So there is not really a problem with "not being able to set class styles". The code is working as expected, when you add the `display: 'block'` to the `makeStyles` it will be applied to the `a` tag in both the **dev** and **prod** environment.

@keyy123 about the link colors I think it is fine to leave it like this. Our links all have different colors, so if we want to change it we'll probably do it by overriding the style.

RubenSmn (commented):
If you could move the **inline** style to the **class style** then I think the issue will be solved

src/components/Home.js
```diff
@@ -1,3 +1,4 @@
+import { Brightness1 } from '@mui/icons-material';
```
RubenSmn, line 1: Could you remove this unused import

src/components/Link.js
```diff
@@ -1,16 +1,22 @@
 /* eslint-disable jsx-a11y/anchor-has-content */
 /* eslint-disable react/display-name */
+import { Block } from '@mui/icons-material';
```
RubenSmn, line 3: Same unused here

RubenSmn (changes requested):

RubenSmn (approved):
Awesome!

RubenSmn:
@keyy123 Could you resolve the merge conflict?

keyy123:
@RubenSmn 😁👌

dadiorchen:
Because we discard the issue #830 so let's close this PR for now.

PR #1274, opened 2022-12-09
refactor: rename test files 

ojafero:
# Description

 Renamed test files to follow convention

# Checklist:


dadiorchen:
Sorry for the delay, I think it is hard to merge this now, someone might has been modified things, let's remain the same for the time being.

PR #1291, opened 2022-12-17
fix: change link color with mode

ojafero:
# Description

Link color will now change with theme mode.

## Type of change

# Checklist:

### DIFF
--- a/src/pages/500.js
+++ b/src/pages/500.js
@@ -1,5 +1,5 @@
-import { useTheme } from '@emotion/react';
 import { Link } from '@mui/material';
+import { useTheme } from '@mui/material/styles';
 import HeadTag from 'components/HeadTag';
 
 export default function somethingWentWrong() {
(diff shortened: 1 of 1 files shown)


src/pages/500.js
```diff
@@ -1,3 +1,4 @@
+import { useTheme } from '@emotion/react';
```
RubenSmn, line 1:
```js
import { useTheme } from '@mui/material/styles';
```

RubenSmn (changes requested):

RubenSmn (changes requested):
I think that an even better approach is to only use the `Link` component with a `sx` prop where we get the theme from. That way we don't need to use inline-styles and don't have a `a` tag in a `a` tag

ojafero:
@RubenSmn That arrangement is the recommended approach https://nextjs.org/docs/upgrading#link-component

RubenSmn:
Ah I see, but we're using the `Link` from `@mui/material` @ojafero

RubenSmn (approved):

PR #1296, opened 2022-12-18
fix: invalid image url

ojafero:
# Description

Show placeholder image when user enters invalid link.

## Type of change

## Screenshots

|  |  |

# Checklist:


RubenSmn (commented):
Maybe we should also provide some feedback to the user when there is an error

dadiorchen:
This is pretty smart, As Rubbem said, before we do this, we'd better define a good UX and error message, now, I will close this PR first, and back to this feature when we give a definition on this, currently, the CWM is in progress, so there is some uncertainty that makes it difficult to do so.

PR #1322, opened 2022-12-23
fix: show date error on expand screen

ojafero:
# Description

No error is shown when the expand button is clicked and an invalid date range is selected on the timeline. This change fixes it.

## Type of change

# Checklist:

### DIFF
--- a/src/components/Timeline.js
+++ b/src/components/Timeline.js
@@ -120,7 +120,17 @@ function Timeline() {
     isGlobalPage && (
       <>
         {isError && (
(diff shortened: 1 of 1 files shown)


RubenSmn (changes requested):
To keep the layout more consistent I think it is better to use `20px` since the other components are using that value as well.

RubenSmn (changes requested):
Looks great but I have found one issue with vertical alignment of the icon of the alert

image

It seems like it is missing a `align-items: center` for the icon container

RubenSmn (approved):

Issue #1071, opened 2022-10-10
Define our repository configuration for dev & relase workflow


dadiorchen:
Please help us to set up/define the workflow on GitHub and release.

Goal:

The github/git configuration for different branch/feature, and consider the deployment and test/dev of this branch, feature.

Requirments:

1. A workflow to be able to develop on different features/versions.
  So we can work on different features at the same time.
3. A rule to handle the semantic version number, like: v1.x v2.x, beta, gamma, for version.
  When we working on different feature, we need to consider the version number, because we are using sementic release (https://github.com/semantic-release/semantic-release) so how do we count the version number when we: working on feature A, B, and maintain the `main` branch at the same time.
5. A way to release and test and really use it (in prod) for different feature/version. 
  We need to consider how to deploy the branches to environments, to deliver the version to the tester, and end-user( it is possible that we want to give multiple version to prod and to be used by our customer/end-user, for example, we want to give a version to part of our users to try/test the new version. Or, we want to maintain the old version to our old user for some legacy problem.

dadiorchen:
done

PR #1349, opened 2023-01-17
feat: connect backend config

RubenSmn:
# Description

- can get config from map-config-api
- currently only `navbar` config implemented

## Type of change

# Checklist:


RubenSmn:
@dadiorchen what do you think of this approach? We need to fetch the config only once when the initial page loads. Then we can pass it down as props.

src/pages/_app.js
```diff
@@ -212,7 +216,21 @@ TreetrackerApp.getInitialProps = async (context) => {
...
+  const mapConfig = await mapConfigRequest.json();
+  const config =
+    mapConfig.find((item) => item.name === 'testing-config')?.data ||
+    defaultConfig;
+
+  return {
+    props,
+    config,
+    device,
+  };
```
dadiorchen, line 233: Do you think React context is a better way, so:
1. We don't need to pass through the config
2. The config is an infrequent change variable (99% of cases), so context fits this case.

Aslo, because we are in `beta` rather a dedicated branch, maybe we need a switcher (like env var) to open/close this feature, I mean, loading global config for the webmap, so we can close this on prod at this stage

RubenSmn: Yes you're right, using context is a better solution. I'll update the functionality to only run on DEVELOPMENT

dadiorchen: If you can find someone who can help us on this devops job then we will not be worried about this kind of thin. https://github.com/Greenstand/treetracker-web-map-client/issues/1071

RubenSmn:
@dadiorchen I updated the fetch to the server to only run on **DEV** and created a context for the config which only gets set once. Let me know what you think!

RubenSmn:
To fix the tests, do I need to add a nock route to this file **./cypress/tests/integration/nockRoutes.js**?

dadiorchen:
@RubenSmn that's an option, or we can close the config fetching by env var as we discussed? so we don't need to care about this?

src/pages/_app.js
```diff
@@ -212,7 +214,25 @@ TreetrackerApp.getInitialProps = async (context) => {
 
   const device = userAgentFromString(userAgent)?.device.type || 'desktop';
 
-  return { props, device };
+  let config = defaultConfig;
+  if (!process.env.NEXT_PUBLIC_SERVER_CONFIG_DISABLED) {
```
dadiorchen, line 218: @RubenSmn are you sure this works for `=false` ? I mean, does next convert the string `false` to bool?

RubenSmn: You're right thanks for pointing this out!

dadiorchen:
@RubenSmn are you talking about the error in CI? this one?

```
FetchError: The following error originated from your application code, not from Cypress.

  > request to https://dev-k8s.treetracker.org/map_config/config failed, reason: Nock: No match for request {
  "method": "GET",
  "url": "https://dev-k8s.treetracker.org/map_config/config",
  "headers": {
    "accept": [
      "*/*"
    ],
    "user-agent": [
      "node-fetch/1.0 (+https://github.com/bitinn/node-fetch)"
    ],
    "accept-encoding": [
      "gzip,deflate"
    ]
  }
}
```

dadiorchen:
@RubenSmn I think the unit test is irrelevant to your code, other devs ran into it too

PR #1350, opened 2023-01-17
Refactor: cwm theme editor

RubenSmn:
# Description

- refactor: reduce rerenders on `ColorInput`
- refactor: reduce rerenders on `ColorThumbnail`
- fix: double if statement case
- refactor: create new context for setters
- refactor: reduce rerenders on `TypographyInput`
- refactor: reduce rerenders on `TypographyThumbnail`
- refactor: extract custom hook to deal with default values for reset logic

## Type of change

- [x] Refactor

# How Has This Been Tested?

- [x] Cypress component tests

# Checklist:

### DIFF
--- a/src/components/playground/TypographyInput.js
+++ b/src/components/playground/TypographyInput.js
@@ -36,7 +36,7 @@ function FontFamilyWeightElm({ label, path, fontValue }) {
   const defaultValue = useDefaultValue(value);
 
   useEffect(() => {
-    setValue(defaultValue.current.toString());
+    setValue(defaultValue || '');
     setError('');
   }, [fontValue, defaultValue]);
 
--- a/src/components/playground/TypographyInput.js
+++ b/src/components/playground/TypographyInput.js
@@ -18,6 +18,7 @@ import {
   usePlaygroundFonts,
   usePlaygroundTheme,
 } from 'hooks/contextHooks';
+import { useDefaultValue } from 'hooks/cwmHooks';
 import {
   predefinedFonts as defaultFonts,
   propRules,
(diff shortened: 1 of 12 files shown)


dadiorchen:
Awsome!

src/components/playground/TypographyInput.js
```diff
@@ -224,15 +224,11 @@ function FontFamily(props) {
...
   const [isValid, setValid] = useState(true);
-  const [defaultValue] = useState(initialValue);
-
-  useEffect(() => {
-    setValue(initialValue);
-  }, [initialValue]);
+  const defaultValue = useDefaultValue(value);
 
   const resetTypography = () => {
     setPropByPath(path, defaultValue);
```
jeremydthomas, line 231: @RubenSmn all of the resetTypography's need to have the .current added to defaultValue or they dont work to reset. There are 3 of them. like this
``` 
 const resetTypography = () => {
    setValue(defaultValue.current);
    setPropByPath(path, defaultValue.current);
  };
```

awesome job on all of this so far though, have a few more individual commits to look at.

jeremydthomas (changes requested):
Everything but that one comment I have there on TypographyInput.js looks really good to me. Awesome job!

src/components/playground/TypographyInput.js
RubenSmn: Thanks I did a lot of testing on the `Color` side of the components. But forgot the `Typography` resets! This gives me the idea that the `.current` is weird to have since the hooks should just return the `value` :smile: .

jeremydthomas (commented):
so everything that I found is fixed, however I did notice that the global reset everything button doesnt reset the words back to default, it does actually set everything back to default but not the correct words under font family, font weight etc. This does this in production right now too so I dont know if you want to fix in this PR or I can approve this and get an issue put up for it.

RubenSmn:
Aah @jeremydthomas thats a great catch thanks! 

I think that its better if that would be fixed in a seperate PR to keep things organized 👍

jeremydthomas (approved):

jeremydthomas:
> Aah @jeremydthomas thats a great catch thanks!
> 
> I think that its better if that would be fixed in a seperate PR to keep things organized 👍

Sounds good to me @RubenSmn

PR #1387, opened 2023-02-09
feat(cwm): added prototype of a global announcement MUI snackbar

### LINKED ISSUE #1249: Implement simple overlay notification
The webmap.config would be the best place to have global announcements. A simple pop up panel with a configurable message to announce downtimes, upgrades or other important messages.

andrewsu31098:
# Description

Added in GlobalNotification component that displays a global announcement through a MUI Snackbar. 
Snackbar values dependent on values in the Webmap Config. 
NOTE:  This is just a prototype because I think config will change in the future, meaning this implementation will need changes too.

Fixes #1249 

## Type of change

## Screenshots

**Before**
before

**After**
after

# Checklist:

### DIFF
--- a/src/context/configContext.js
+++ b/src/context/configContext.js
@@ -6,7 +6,7 @@ import {
   useEffect,
 } from 'react';
 import { useLocalStorage } from 'hooks/globalHooks';
-import configReducer from '../models/config.reducer';
+import configReducer from 'models/config.reducer';
 
 const ConfigContext = createContext(null);
 
@@ -47,6 +47,12 @@ export const initialState = {
       lon: '',
       zoom: '',
     },
+  },
+  globalMessage: {
+    visible: false,
+    severity: 'info',
+    message:
+      ' Scheduled maintenance from 3:00am PST to 6:00am PST. (This is an example)',
   },
 };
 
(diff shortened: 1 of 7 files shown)


RubenSmn (changes requested):
This looks really great @andrewsu31098! I do see some styling issues with the **info** and **close** icon not being vertically aligned in the center. Could you try to fix these?

RubenSmn:
When simulating mobile on the browser the Alert gets removed from the DOM

andrewsu31098:
Gotcha! Will work on it 👍

src/context/configContext.js
```diff
@@ -48,6 +48,12 @@ export const initialState = {
       zoom: '',
     },
   },
+  globalMessage: {
+    visible: false,
+    severity: 'info',
+    message:
+      ' Scheduled maintenance from 3:00am PST to 6:00am PST. (This is an example)',
+  },
```
dadiorchen, line 56: Great job!
I guess this will display on the app, right? Can we build a unit/integration test to demo this rather than put it into the real code?

dadiorchen (changes requested):

src/context/configContext.js
dadiorchen: @andrewsu31098 my suggestion is: build a Notification.cy.js unit test for demonstration, in this way, we can cover the code by test, also, it is a handbook for people to use in the future, integrating into the layout is fine and easy, we can do it when needed.

RubenSmn: @andrewsu31098 do you need any help or guidance with this?

dadiorchen:
I will close this PR now, but the code here is a good reference if someone can continue this task.

Issue #1472, opened 2023-03-14
Change the javascript in Github action to appropriate language


dadiorchen:
https://github.com/Greenstand/treetracker-web-map-client/blob/253ad26777d7010a2c9b5c2c1af5bcd38991bd16/.github/workflows/dev-build-release.yaml#L30

This is for a temporary purpose in Javascript, is there any other better solution?

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- For design guidelines checkout our [Figma design](https://www.figma.com/file/XdYFdjlsHvxehlrkPVYq0l/Greenstand-Webmap).
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.
- More issues in this project: https://github.com/orgs/Greenstand/projects/26

diegovanelli:
Hi!

PR #1492, opened 2023-03-20
chore: change the javascript in Github action to appropriate language

diegovanelli:
# Description

Change the javascript in Github action to appropriate language

Fixes https://github.com/Greenstand/treetracker-web-map-client/issues/1472

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/.github/workflows/dev-build-release.yaml
+++ b/.github/workflows/dev-build-release.yaml
@@ -34,7 +34,7 @@ jobs:
               RELEASE_NEEDED=true
               break
             fi
-          done < <(jq -r '.branches[].name' ./.releaserc.json)
+          done < <(jq -r '.branches[].name' .releaserc.json)
           echo "::set-output name=RELEASE_NEEDED::$RELEASE_NEEDED"
         id: release_needed
       - name: Print rn
--- a/.github/workflows/dev-build-release.yaml
+++ b/.github/workflows/dev-build-release.yaml
@@ -34,7 +34,7 @@ jobs:
               RELEASE_NEEDED=true
               break
             fi
-          done < <(jq -r '.branches[].name' .releaserc.json)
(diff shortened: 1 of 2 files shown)


dadiorchen:
@diegovanelli great, did you test it?

diegovanelli:
Hello!
Yes, I did.
But I saw a problem in the .releaserc.json file which has different pattern in the branch names. One option is to standardize as an object and set the name of each branch.

What do you think?

dadiorchen:
@diegovanelli do you mean these two: {master}, and {name: 'cwm', channel}, this the default patten in semantic release, I hope we can support this standard format, by ignoring the extra properties like channel, it means the same name as the given one.

.github/workflows/dev-build-release.yaml
```diff
@@ -27,7 +27,15 @@ jobs:
...
-        run: echo ::set-output name=RELEASE_NEEDED::$(echo `node -e 'console.log(require("./.releaserc.json").branches.map(e => e.name || e).reduce((a,c) => a || c === process.env.GITHUB_REF_NAME, false))'`)
+        run: |
+          RELEASE_NEEDED=false
+          while read -r branch; do
+            if [ "$branch" == "${GITHUB_REF_NAME}" ]; then
+              RELEASE_NEEDED=true
+              break
+            fi
+          done < <(jq -r '.branches[].name' '.releaserc.json')
+          echo "::set-output name=RELEASE_NEEDED::$RELEASE_NEEDED"
```
dadiorchen, line 38: I think bash isn't a better way comparing to JS, it is also problem prone, maybe we can find some Github Action to help on this, a component/plugin

PR #1521, opened 2023-04-23
feat: add time elements in wallet page

Harshalyadav:
# Description

Added Time element on the wallet page and Wrap into fragments for both Mobile and PC view.


RubenSmn:
Have been fixed by #1524

PR #1540, opened 2023-05-01
fix(wallet): default list style showing

RubenSmn:
# Description

Removes default list style which was introduced by changing the HTML structure.

## Type of change

# Checklist:


dadiorchen:
@RubenSmn thank you

PR #1541, opened 2023-05-02
feat: organization page made species plants in list

Harshalyadav:
# Description

Species plants are shown in the list

# Checklist:

### DIFF
--- a/src/pages/organizations/[organizationid].js
+++ b/src/pages/organizations/[organizationid].js
@@ -332,12 +332,14 @@ export default function Organization(props) {
             </Grid>
           </Grid>
           <Box
+            component="ul"
             sx={{
               px: [0, 6],
               display: isPlanterTab ? 'block' : 'none',
+              listStyle: 'none',
             }}
           >
-            <Box sx={{ mt: [0, 22] }}>
+            <Box component="li" sx={{ mt: [0, 22] }}>
               <CustomWorldMap
                 totalTrees={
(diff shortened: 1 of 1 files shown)


src/pages/organizations/[organizationid].js
```diff
@@ -357,17 +357,20 @@ export default function Organization(props) {
...
               }}
             >
               {organization?.species?.species?.map((s) => (
                 <React.Fragment key={s.name}>
-                  <TreeSpeciesCard
-                    name={s.name}
-                    subTitle={s.desc || '---'}
-                    count={s.total}
-                  />
+                  <li key={s.id}>
```
Comment (RubenSmn) [inline src/pages/organizations/[organizationid].js:367]: No need for a key here since this is not the _top level_ component in the loop.

RubenSmn (changes requested):
We should also convert the [_hired planters_](https://github.com/Greenstand/treetracker-web-map-client/pull/1541/files#diff-1160c2738fc888e43168fd35336e393916c22bb258e8c174a4bfeee63d0fd7c3R385-R431) section to a list. We can easily use the `component` prop for this.

Also we might need to fix the default styling from the lists. Take a look at this [PR](https://github.com/Greenstand/treetracker-web-map-client/pull/1540)

src/pages/organizations/[organizationid].js
Comment (Harshalyadav) [reply src/pages/organizations/[organizationid].js:367]: Ok, I see

RubenSmn:
Resolved by #1559

Issue #1546, opened 2023-05-05
Display 500 error page inappropriate.


dadiorchen:
When visiting page, if the resource (wallet name or so) doesn't exist, we shouldn't display 500 page, please display 404 instead.

for example:
https://map.treetracker.org/wallets/8a0ce006-bb2d-40f3-8005-8c3080dca6d3

<img width="1416" alt="image" src="https://user-images.githubusercontent.com/5744708/236395326-454f0550-2c9f-4982-96bb-a21b8e0fba5d.png">

This is a good way to reproduce the problem: https://github.com/Greenstand/treetracker-web-map-client#how-to-connect-to-production-data-locally

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- For design guidelines checkout our [Figma design](https://www.figma.com/file/XdYFdjlsHvxehlrkPVYq0l/Greenstand-Webmap).
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.
- More issues in this project: https://github.com/orgs/Greenstand/projects/26

Sdaewa:
Hi there, 
if I get this issue assigned, where should I find the 404 design page or that would be up to me for design?
thanks

dadiorchen:
You don’t need to design anything, 404 and 500 page is there already, the issue is about how to jump to the correct error page.

Dadior Chen

在 2023年5月5日，16:55，Luis ***@***.***> 写道：

﻿

Hi there,
if I get this issue assigned, where should I find the 404 design page or that would be up to me for design?
thanks

Sdaewa:
okay I understand now, I will work on it then, thanks

Harshalyadav:
@dadiorchen please check my pr.

dadiorchen:
@Sdaewa seems @Harshalyadav have solved this, can you help review this PR?

dadiorchen:
This is an issue open for everyone

dadiorchen:
@Harshalyadav I remember your solution is good, but because the chaos of the branches so we closed it, can you pick your previous code and do a new PR again?

yasiru98:
[...]

PR #1550, opened 2023-05-05
Walletid not match redirect 404 #1546

Harshalyadav:
# Description

If the wallet id is wrong or does not match with the database wallet id then it will show a 404 page.


RubenSmn:
I think you should open against the `beta` branch instead of the `main` branch, correct @dadiorchen ?

dadiorchen:
@Harshalyadav thank you for you contribution, really sorry we encounter some difficulty in maintaining the branches, now we have merged `beta` to `main`, but I don't know why there still are some many differences in this PR, I thought it should have gone if we merge it, so can you take a took at this, if there isn't better solution, I have to require you to checkout the main branch again, folk another branch then raise PR.
If you commit is valid, you can do this by `git cherry-pick [your commit]` to fetch that commit to your new branch

dadiorchen:
By the way please check if those two other PR by you is valid (their name is similar)

Harshalyadav:
ok, I check those PR.

dadiorchen:
@Harshalyadav sorry, can you raise a new pr that just includes your solution (I guess is the line changing 500 to 404) so I can easily merge it? I will close this one, sorry for all these disturbing!

Harshalyadav:
Ok, I'm creating new PR.

PR #1598, opened 2023-05-25
fix: upgrade cypress action to try to fix ci problem

dadiorchen:


jeremydthomas:
is this the fix for 
API CI for New Pull Requests / Install dependencies (pull_request) Failing after 6s?

dadiorchen:
Thanks for looking at this, we got some really difficulty with the cypress test, I raise a question on their side: https://github.com/cypress-io/github-action/issues/922

dadiorchen:
@jeremydthomas this is my fault, I have solved this, it is not the problem with cypress action.

jeremydthomas:
> @jeremydthomas this is my fault, I have solved this, it is not the problem with cypress action.

Great to hear that you fixed it

PR #1619, opened 2023-05-30
refactor(cwm): generate inputs from config file

### LINKED ISSUE #1604: CWM theme editor refactor
### Problem

Currently we're using the [theme from the themeContext](https://github.com/Greenstand/treetracker-web-map-client/blob/main/src/context/themeContext.js) to render our inputs for the editor this does not scale too well.

We currently have 3 types of inputs, 'text', 'select', 'color'. How we determine which input to render is not clear :cry:. The way the data flows from the `theme` constant to the `input` being chosen is hard to follow.

### Approach

We're going to take a more config based approach. The following code snippet is from [this project](https://github.com/RubenSmn/spotify-banner-creator/blob/main/src/constants/config.ts) which I build almost a year ago.

```javascript
const config = {
  fontSize: {
    displayText: 'Size',
    input: PROPTYPES.SLIDER, // 'slider'
    min: 3,
    max: 5,
    step: 0.2,
    defaultValue: 4,
    unit: 'rem',
  },
};
```

As you can see the **idea** is that we can add a property to the `config`, define the requirements for that specific property, such as what kind of input it needs to be e.g. 'slider', 'text', the additional values according to that input type.

---

This new flow will improve developer experience by being able to configure the sections e.g. colors, typography in a single config file and improve the UI/UX for the end user by introducing more straight forward inputs based on the value type.

RubenSmn:
# Description

- Generate inputs from config file instead of theme object.
- Updating the theme works in the same way.
- Allows for easier improvements and customization for input UI (DX).

Fixes #1604

## Type of change

- [x] Refactor

# How Has This Been Tested?

- [x] Cypress component tests

# Checklist:


dadiorchen:
@RubenSmn please hold on to this PR, I need to discuss with you something, maybe in a meeting

dadiorchen:
Can I run this locally?

RubenSmn:
Yes you can @dadiorchen

PR #1633, opened 2023-06-04
Feat: add new badge on [captureid] page to check tree match status

### LINKED ISSUE #1586: Add `tree-matched` badge on the `/captures/{id}` page
UX: to the right of the `token not issued` badge, add a new badge, which:
* If the `capture` object's `tree_id` is not null, then the badge shows: `tree matched` with primary `green` color, and by clicking the badge, the page jumps to: `/trees/{tree_id}`
* If the `tree_id` is null, then the badge shows: `waiting for tree match` in grey color

http://localhost:3000/v2/captures/0c2d6e3a-d02a-4edf-bcdc-893e1f7a4817

image

---
some hints:
- Dev guides for `v2` feature: https://docs.greenstand.org/web-map/v2-web-map
- If you want to join the Slack community (some resources need the community member's permission), please leave your email address.

khalatevarun:
# Description

* Add tree-matched batch
* The batch will be displayed only when showTreeMatchStatus is passed since it is also used on treeid page where we don't need to show.

Fixes #1586 

## Type of change

## Screenshots

### When tree.id is undefined or does not exist
image

### When tree.id exists
image

# How Has This Been Tested?

- [ ] Cypress integration
- [ ] Cypress component tests
- [ ] Jest unit tests

# Checklist:


khalatevarun:
@dadiorchen please have a look, let me know if my approach is good or needs to changed.

khalatevarun:
@RubenSmn can you have a look at it once?

khalatevarun:
@dadiorchen can we merge this PR or does it need any changes or improvement?

dadiorchen:
@khalatevarun this Badge component is getting unnecessarily complicated now, do you think we can do a refactor to make it more reusable? The problem is that the component knows too much, like: `tokenId`, if we can remove all those things, maybe we can make it simple and easy to reuse.

RubenSmn:
Ye I was thinking about this too, but didn't know if we wanted to introduce a new feature and a refactor in the same PR.

As for the refactor I think it would be nice to have separate badges, say a `VerifiedBadge` and a `MatchStatusBadge` which we can use individually instead of one `Badges` component. I think this makes more sense in terms of the component hierarchy and readability.

dadiorchen:
@khalatevarun I hope you can do this refactor in the `main` branch and sync code to the v2 later, because accurately, this doesn't belong to `v2` and would cause conflict in the future if we do it on v2.

dadiorchen:
@RubenSmn is it possible to create a general `badge` component, and leave the client to decide the details? so the `badge` just communicates with the invoker by: text, color, icon?

RubenSmn (commented):
@khalatevarun I think this solution introduces less readable code. I think we should explore something like a `variant` prop to make a generic `Badge` component.

Usage of that component could be something like this:

```javascript
<Box>
  <Badge variant="verified">
    {tree?.approved ? "Waiting for verification" : "Verified"}
  </Badge>
</Box>
```

And then in the `Badge` component we do something like this:

```javascript
const badgeProps = {
  verified: { styles, icon, etc },
  "awaiting-approval": {},
  "other-variants": {},
};

function Badge({ variant, ...rest }) {
  const props = badgeProps[variant];

  return <Chip size="small" {...props} {...rest} />;
}
```

So we make one generic `Badge` component instead of specific badges. This allows also for easy scaling since we can simply add a new `variant`.

khalatevarun:
@RubenSmn i think one of the reasons why we build reusable component is too have the same design of the component across the application, if we expose too many control to the parent component using it then there is not sense in making it reusable, we can rather directly go with Chip instead of wrapping it with HOC.

In PR - 1643, I have tried to expose only event handlers and some variables to change status keeping other attributes private to the Badges component, does it make sense?😅

RubenSmn:
I have to disagree here @khalatevarun, we provide a singe prop `variant` which defines a preset of properties which are needed for a single **variant**, this abstracts a lot of things like style and icons. We allow for a **opt-in** to add more props as needed, say a `onClick` event.

dadiorchen:
I feel a styled `chip` is good enough, I prefer to give back some responsibility to parents to describe things, but I admit this is a trade-off

khalatevarun:
Closing this PR since no change are required as the PR to fix this issue was merged to main and then to v2

PR #1643, opened 2023-06-16
refactor the badge component for better reusability

khalatevarun:
# Description

* Updated badges component to take an array of data as props
* commented the test code temporarily, will raise a separate PR for the same

Fixes # (issue)

## Type of change

## Screenshots

# How Has This Been Tested?

- [ ] Cypress integration
- [ ] Cypress component tests
- [ ] Jest unit tests

# Checklist:


src/components/Badges/index.js
```diff
@@ -1,18 +1,30 @@
 import VerifiedBadge from '../VerifiedBadge';
 
-function Badges({ tokenId, verified }) {
+function Badges({ content }) {
```
dadiorchen, line 3: @khalatevarun my thoughts:
1. Badges vs Badge, the later is more flexible and reusable, say, if we want to put the Badges in another place, maybe the elements wrap the Badges will change, so it becomes harder for Badges to adapt to that change, the more simple of the component, the better to use.
2. content vs explicit parameter list (color, disabled), the latter is more safe and clear to the user of this component, interface should be a contract with user rather than a blurred object that can be anything in the future.

src/components/Badges/index.js
```diff
@@ -1,18 +1,30 @@
...
+          badgeName,
+          onClick = null,
+          disabled = false,
+        } = data;
+        return (
+          <VerifiedBadge
+            // eslint-disable-next-line react/no-array-index-key
+            key={index}
+            color={color}
+            verified={verified}
```
dadiorchen, line 20: verified isn't a language that the general component should know, what if we want to build a badge for `matched` rather than `verified` ? with color and icon is good enough, or we can use icon name like: leading-icon='verified', 'matched', 'null'

khalatevarun:
Your thoughts make sense @dadiorchen 
But i have a doubt now, do we really need to make them reusable then? We can simply use the Chip component then?😅😅😅

khalatevarun:
I have directly used the Chip component in the Badges component in the latest commit, this is what I could come up with @dadiorchen

src/pages/trees/[treeid].js
```diff
@@ -284,7 +297,7 @@ export default function Tree({
                   mt: 2,
                 }}
               >
-                <Badges tokenId={tree.token_id} verified={tree.approved} />
+                <Badges content={badgesContent} />
```
Comment (dadiorchen) [inline src/pages/trees/[treeid].js:300]: @khalatevarun can change the interface to something like this:
<Badge color="primary" name="xxxx" />
Don't use array

dadiorchen (changes requested):

dadiorchen:
> I have directly used the Chip component in the Badges component in the latest commit, this is what I could come up with @dadiorchen

Yea, if we can use raw component, why do we need to use extra ones? Maybe we just need to change the style of the chip is good enough for us? like this: https://mui.com/material-ui/react-button/#customization

src/pages/trees/[treeid].js
Comment (khalatevarun) [reply src/pages/trees/[treeid].js:300]: okay

Comment (RubenSmn) [reply src/pages/trees/[treeid].js:300]: I made some suggestions on #1643

Comment (dadiorchen) [reply src/pages/trees/[treeid].js:300]: @khalatevarun I think the array one is harder to read. The code is scattered in places. And the array component is more complicated than single component, and hard to reuse

Comment (khalatevarun) [reply src/pages/trees/[treeid].js:300]: Okay, will simply render Bage one after the another

Comment (RubenSmn) [reply src/pages/trees/[treeid].js:300]: I suggested some things [here](https://github.com/Greenstand/treetracker-web-map-client/pull/1633#issuecomment-1598464948) @dadiorchen

khalatevarun:
@dadiorchen does this look good now?

dadiorchen:
@khalatevarun this looks great! Yea, please go ahead removing things.

khalatevarun:
Done with the cleaning @dadiorchen

dadiorchen:
@khalatevarun thank you for taking some much effort and patience to finish this, this is big improvement for us

dadiorchen:
I will merge to `main` and then to `v2` so you can continue that issue.

khalatevarun:
yes, please let me know once its merged to v2 so that I can continue with that issue @dadiorchen

dadiorchen:
@khalatevarun I have merged to v2

PR #1660, opened 2023-06-28
Update utils.js

Harshalyadav:
# Description

Walletid not match redirect 404 #1546

Fixes # (issue)

#1546

### DIFF
--- a/src/models/utils.js
+++ b/src/models/utils.js
@@ -231,7 +231,7 @@ const wrapper = (callback) => (params) =>
     if (e.response?.status === 404) return { notFound: true };
     return {
       redirect: {
-        destination: '/500',
+        destination: '/404',
         permanent: false,
       },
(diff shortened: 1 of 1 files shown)


src/models/utils.js
```diff
@@ -231,7 +231,7 @@ const wrapper = (callback) => (params) =>
     if (e.response?.status === 404) return { notFound: true };
     return {
       redirect: {
-        destination: '/500',
+        destination: '/404',
```
dadiorchen, line 234: When I re-think this change, I think it is problematic, the previous logic is:
1. If the status code is 404, then return notFound, which means 404 page.
2. If any other case, return 500, means, unknown internal error
So I think we shouldn't make all error 404, that's confusing.

Back the the issue, the URL return 500 but 404, I thought this is because of some 500 error in the server side, but it is 404, so we need to debug the specific URL to see what happened, and handle that error then throw 404 to the wrapper.

PR #1661, opened 2023-06-29
fix: keep tree_id parameters when bounds are updated

### LINKED ISSUE #1556: When update the url with `bounds`, we should remain all other existing ones (parameters)
Check the video:
https://dadior.s3.ap-northeast-1.amazonaws.com/Screen%20Recording%202023-05-10%20at%203.29.45%20PM.mp4

**Describe the bug**

**To Reproduce**
(we recommend record a screenshot to show the case, like using http://loom.com to record)

**Expected behavior**

**Screenshots**

\*\* Node.js version
(The node version on your computer to run the project)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

henlam1:
# Description
- I've tested this locally by selecting the wallet at http://localhost:3000/wallets/tendry@bondy.earth?tree_id=5413738
- I used a for loop to add query parameters into the router query
- Since `router.query.tree_id` is a string and `result.query.tree_id` was a number, I casted `tree_id` to string because it was an eslint requirement that they're the same type.
- I added a new test case in pathResolver.test.js

Fixes #1556 

## Type of change

## Screenshots

# How Has This Been Tested?

- [ ] Cypress integration
- [ ] Cypress component tests
- [x] Jest unit tests

# Checklist:


src/models/pathResolver.js
```diff
@@ -93,6 +93,9 @@ function updatePathWhenMapMoveEnd(location, map, router) {
   }${router.query.embed ? `&embed=true` : ''}`;
   if (router.query.tree_id) {
     result += `&tree_id=${router.query.tree_id}`;
+  } else if (router.query.walletid) {
+    const tree_id = router.asPath.match(/tree_id=(\d+)/)[1];
+    result += `&tree_id=${tree_id}`;
   }
```
dadiorchen, line 99: @henlam1 thanks for taking time to read code and do a PR, this is not so easy, given the fact that our complicated code.

Your code seems a way to handle this spesicifc case: walletid, but ideally, it would be better if we can handle this in a more general way, that means, we don't need to check the walletid, instead, we can detect all the parameter and put it into the path, does this make sense?

dadiorchen: Bye the way, we use jest test to unit test this model, so please run it locally to make should it's okay, also, if you can write some unit test to cover your new code, it would be great, like, pass in the URL of our case (URL with wallet and treeid), and check the result in your test

dadiorchen: https://github.com/Greenstand/treetracker-web-map-client/blob/main/src/models/pathResolver.test.js this is the test file

henlam1: @dadiorchen thanks for reviewing my PR. 

I agree it's better to handle this in a more general way and this issue is difficult because of the wallet URL structure. 
Selecting a tree produces different URLs depending on the query. For example,

- Planters: `planters/xxxx/trees/xxxx?bounds=`
- Organizations: `organizations/xxxx/trees/xxxx?bounds=`
- Wallets: `wallets/walletId/tokens?tree_id=xxxx` or `wallets/walletId/tokens?bounds=`

I'm not familiar with the wallet routing so I'd like to clarify and understand if the wallet URLs need to use `tree_id=xxxx` instead of `trees/xxxx`

henlam1: And thanks for letting me know about the unit tests

dadiorchen: I can not understand the question, for a general solution, you don't care the specific path (/trees or /wallet) you simply copy all those parameters (query param) into the new one

henlam1: @dadiorchen What I meant to ask is why does /wallet need a query to store the tree id? For /trees and /wallet, the tree id is stored in the pathname of the Location object.

Also I found a more general solution that passed the unit tests. Would greatly appreciate any feedback.

dadiorchen: Got you, yes, that's a special case, for wallet case, we can not use a token id, so we choose to work around by putting tree id here.

dadiorchen:
@henlam1 thank you, can you try run your code locally and check if the bug is gone? you can set this to reproduce the bug: https://github.com/Greenstand/treetracker-web-map-client/tree/cwm#how-to-connect-to-production-data-locally

henlam1:
@dadiorchen glad to help. I'm locally connected to the production data and the bug is gone when I tested. The new code passes all unit tests in pathResolver.test, including the new test case I included.

dadiorchen:
@henlam1 that's great, thank you for your contribution, hope you can help us more!

PR #1681, opened 2023-07-05
Theme from os

Ewan88:
# Description

This change will set the users theme from the OS preferences. If these preferences cannot be accessed then the application will default to the light theme.

The change should not break anything but it's a good idea to test it manually on someone else's machine before merging. To do so you need to delete the previous theme from Application -> Local Storage in dev tools.

## Type of change

# How Has This Been Tested?

- [x] Cypress integration
- [x] Cypress component tests
- [x] Jest unit tests

# Checklist:


RubenSmn (approved):
Great addition @Ewan88 thank you :+1:

PR #1693, opened 2023-07-15
Update README.md

aayushgauba:
I was following through the README and this is something that I thought needed to be corrected


RubenSmn (approved):
Nice catch!

PR #1708, opened 2023-08-31
fix: allow page to navigate back to valid url when bounds change

### LINKED ISSUE #725: Ignore history/back for updating the bounds on url address
Currently, everytime we change the bounds (https://prod-k8s.treetracker.org/web-map-beta/demo/top?bounds=-108.10546875000001,-54.77534585936448,145.01953125000003,59.5343180010956), for example, click the zoom in/out button,  the new bounds will push the browser history, we need to ignore this action, because it make no sense that when people clicking back button on browser, the URL jumps back to previous bounds, instead, we should jump to a previous valid URL. 

https://dadior.s3.ap-northeast-1.amazonaws.com/Screen%20Recording%202022-08-02%20at%2010.42.48%20AM.mp4

QuanGiap:
# Description

I have fixed the website that allows the page to navigate back to the valid URL instead of the previous bounds

Fixes #725 

## Type of change

## Videos

[screen-capture.webm](https://github.com/Greenstand/treetracker-web-map-client/assets/81385914/4226689e-2973-4714-9bac-0fa215603409)

# How Has This Been Tested?

- [x] Cypress integration
- [x] Cypress component tests
- [x] Jest unit tests

# Checklist:

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -171,7 +171,10 @@ function MapComponent() {
           map,
           router,
         );
-        window.history.pushState('treetracker', '', path);
+        // changing the url in history state
+        // this allow website to load page back
+        const historyState = { ...window.history.state, url: path };
+        window.history.replaceState(historyState, '', path);
       });
     }
 


src/components/App.js
```diff
@@ -171,7 +171,10 @@ function MapComponent() {
           map,
           router,
         );
-        window.history.pushState('treetracker', '', path);
+        // changing the url in history state
+        // this allow website to load page back
+        const historyState = { ...window.history.state, url: path };
+        window.history.replaceState(historyState, '', path);
```
dadiorchen, line 177: I think this makes the app can not remember the previous jumps, for the case that doesn't includes bounds, like: /top, /wallets/xxxx, because we always replaceState, please try to jump on dev env is difficult, you set this to connect to prod env: https://github.com/Greenstand/treetracker-web-map-client#how-to-connect-to-production-data-locally

dadiorchen (changes requested):
@QuanGiap thank you for the contribution, I have some suggestion below:

src/components/App.js
QuanGiap: > I think this makes the app can not remember the previous jumps, for the case that doesn't includes bounds, like: /top, /wallets/xxxx, because we always replaceState, please try to jump on dev env is difficult, you set this to connect to prod env: https://github.com/Greenstand/treetracker-web-map-client#how-to-connect-to-production-data-locally

Hi Dadiochen,
Thanks for taking a look at my code and giving instructions on how to connect to prod env, my local is working much better now. 
From what I understand, you said that the website doesn't remember the previous URL that doesn't include the bounds since I use the replaceState. So what is happening is the replaceState only happens when the bounds are changed which is a triggered event 
https://github.com/QuanGiap/treetracker-web-map-client/blob/Ignore-history/back-for-updating-the-bounds-on-url-address-%23725/src/components/App.js#L167
So when you to other pages, it still stores the page you have visited in the history.

dadiorchen:
@QuanGiap you are right, I didn't realize we are in that event handler, thank you! Great job!

And if you are interested in discovering more about Greenstand, please join our community!

PR #1711, opened 2023-09-05
Fix accessibility issues on top page

ojafero:
# Description

-  Added label to slider button so screen readers announce what the button is for
-  Made alt attribute empty for decorative images
-  Made alt attribute for change theme button precise
-  Screen readers announce when light mode or dark mode is turned on

## Type of change

# Checklist:


RubenSmn (approved):
These are some great changes! We should try to make more of these so everyone can use and enjoy the treetracker. Thanks @ojafero

PR #1714, opened 2023-09-07
Fix: replace '\n' with '<br/>' for line breaks in wallet.about

frankienstan:
# Description

Fixes # Fixes an issue where '\\\n' characters in wallet.about content were not rendered correctly as line breaks. Replaced '\\n' with '\<br>' tags to ensure proper line breaks in the rendered HTML content.

## Type of change

### DIFF
--- a/src/pages/wallets/[walletid].js
+++ b/src/pages/wallets/[walletid].js
@@ -374,7 +374,9 @@ export default function Wallet(props) {
           sx={{ mt: [2.5, 5] }}
           variant="body2"
           dangerouslySetInnerHTML={{
-            __html: marked.parse(wallet.about || 'NO DATA YET'),
+            __html: marked.parse(
+              (wallet.about || 'NO DATA YET').replace(/\\n/g, '<br/>'),
+            ),
           }}
         />
         <Divider


src/pages/wallets/[walletid].js
```diff
@@ -374,7 +374,9 @@ export default function Wallet(props) {
           sx={{ mt: [2.5, 5] }}
           variant="body2"
           dangerouslySetInnerHTML={{
-            __html: marked.parse(wallet.about || 'NO DATA YET'),
+            __html: marked.parse(
+              (wallet.about || 'NO DATA YET').replace(/\\n/g, '<br/>'),
```
Comment (dadiorchen) [inline src/pages/wallets/[walletid].js:378]: @frankienstan thank you for your work, because the `about` is in markdown format, so I don't think you should use html tag `<br/>` here, you should replace with newline `\n` and give it to `marked.parse` to generate the content (in html)

BTW, can paste a screenshot of the result?

Comment (frankienstan) [reply src/pages/wallets/[walletid].js:378]: @dadiorchen  Yea sure ,
Before
before
After
after

Comment (dadiorchen) [reply src/pages/wallets/[walletid].js:378]: Yea, the <br/> can do the work of breaking new line, but we don't need to do this by ourselves, we just need to fix the extra `\\n` for the string, and pass it to markdown lib, it will take care of all the work including the newline breaking

Comment (frankienstan) [reply src/pages/wallets/[walletid].js:378]: we can replacing \\n with actual newline characters \n in the wallet.about content before passing it to marked.parse(). This way, marked.parse() will handle the Markdown rendering, including line breaks, as expected.

Comment (dadiorchen) [reply src/pages/wallets/[walletid].js:378]: @frankienstan yes, exactly

Issue #1762, opened 2024-02-24
Use 3rd part action to handle the semantic release


dadiorchen:
Currently, we are using self-managed action to handle the semantic release, majorly, we need to handle this case:
* Multiple branch can working on different feature, and push code to that branch
* The branches can have some channel to distribute, like alpha, beta, which is defined in the semantic release rc file: `.releaserc`

So now we need to try to find some 3rd part lib can handle this and give a stronger support. Here are some candidates:

https://github.com/codfish/semantic-release-action
https://github.com/cycjimmy/semantic-release-action

The goal is setting 3rd part action to do the semantic release, satifys the two points mentioned above.

---

Some hints:

- Please read our [readme](https://github.com/Greenstand/treetracker-web-map-client#treetracker-web) for more information/guide/tutorial.
- For design guidelines checkout our [Figma design](https://www.figma.com/file/XdYFdjlsHvxehlrkPVYq0l/Greenstand-Webmap).
- Here is [an engineering book](https://greenstand.gitbook.io/engineering/) in Greenstand.
- To know more about our organization, visit our [website](https://greenstand.org).
- If you want to join the slack community (some resources need the community member's permission), please leave your email address.

RingoTC:
I can do it. Please assign it to me.

dadiorchen:
Or to try the new 3rd part action, we can start from an easier case: the query API, so to deploy the query API using the 3rd part action, and deploy to different channel.

PR #1771, opened 2024-03-04
update github action & semantic-release for channel distribution

RingoTC:
# Description

[comment]:
Update Github Action configuration and semantic-release configuration for channel distribution. #1762 

- branches leading with feature/ will deploy to the alpha channel
- branches leading with bugfix/ and test/ will deploy to the beta channel
- the main branch will deploy to the prod channel

## Type of change


.releaserc.json
```diff
@@ -21,9 +21,19 @@
       "channel": "next"
     },
     {
-      "name": "demo",
+      "name": "feature/**",
```
dadiorchen, line 24: Does the semantic release accept wide matching? Do you have any document or example for this? I don't know if it does.

RingoTC:
Yes, actually the wide matching is implemented by the GitHub action workflow configuration.

For example, when we make a push with branch name like "feature/test4semantic", then the dev distribution will trigger only.

image

And I think the real purpose of semantic-release is to generate version documentation automatically and push the patch to package management site like npm. So we can only change the workflow configuration to achieve our purpose that select different workflow process depends on the branch name.

dadiorchen:
Yea, github action support the wide matching, but seems the semantic release lib doesn't so it can not trigger the thing for real branch like feature/1

PR #1802, opened 2024-10-24
fix: handle api 404 error to display data properly

### LINKED ISSUE #1801: 404 error on /top page
steps to reproduce:
1. run `npm run dev` on v2 branch
2. click on the green "Let's Find a Tree" button on the home page
image
image

---
some hints:
- Please read this before you start: https://docs.greenstand.org/web-map/v2-web-map
- If you want to join the Slack community (some resources need the community member's permission), please leave your email address.

yunchipang:
# Description

Fixes #1801

⛔️ Cause: when populating static props (captures, countries, growers, organizations, wallets) in `top.js`, if any of the 5 items got a 404 upon API request, the page renders a 404 not found because the field was not properly populated.

🧩 Solution: consolidate all API request functions into `src/models/api.js`, including error handling. When an API request failed, the function returns an empty array.

📝 Note: For now, API requests to `captures` and `growers` return 404, so the below error persists, and the frontend is only displaying `countries`, `organizations`, and `wallets` (please refer to the screenshot in the next section). I believe the investigation should continue on [treetracker-query-api](https://github.com/Greenstand/treetracker-query-api) in terms of fetching captures and featured growers. I've checked `/captures` and `/growers/featured` endpoints, neither of them are working.

image

## Type of change

## Screenshots

| image | image |

# How Has This Been Tested?

- [x] Cypress integration
- [ ] Cypress component tests
- [ ] Jest unit tests

# Checklist:

- [] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/models/api.js
+++ b/src/models/api.js
@@ -10,10 +10,10 @@ export async function getFeaturedGrowers() {
     const res = await axios.get(url);
     const data = await res.data;
     log.warn('url:', url, 'took:', Date.now() - begin);
-    return data.grower_accounts;
+    return data.grower_accounts || [];
   } catch (err) {
-    log.error(err.message);
-    throw err;
+    log.error('Error in getFeaturedGrowers:', err.message);
+    return [];
   }
 }
 
@@ -24,10 +24,10 @@ export async function getCaptures() {
     const res = await axios.get(url);
     const data = await res.data;
     log.warn('url:', url, 'took:', Date.now() - begin);
-    return data.captures;
+    return data.captures || [];
   } catch (err) {
-    log.error(err.message);
-    throw err;
+    log.error('Error in getCaptures:', err.message);
+    return [];
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/models/api.js
```diff
@@ -24,10 +24,10 @@ export async function getCaptures() {
     const res = await axios.get(url);
     const data = await res.data;
     log.warn('url:', url, 'took:', Date.now() - begin);
-    return data.captures;
+    return data.captures || [];
   } catch (err) {
-    log.error(err.message);
-    throw err;
+    log.error('Error in getCaptures:', err.message);
+    return [];
```
dadiorchen, line 30: @yunchipang why you disable the throwing here?

yunchipang: because if it throws an error, `getCaptures()` will not return anything, then `captures` in static props gets undefined, and the entire top page just shows 404.

for now, to inform the error i'm just logging it. would you like this to be handled in some other way? my purpose is to have some props display on the page even when some other props were not populated properly.

dadiorchen: I'm not sure, let's merge your code first, and we can continue, we can back to this if there are some problem. 

BTW, I also don't like the try-catch way to handle error, I prefer to use Promise function to do it:

promise
.then(goodHandler)
.catch(badHandler)

dadiorchen (approved):

dadiorchen:
@yunchipang thanks for you contribution!

